PR #850, opened 2023-10-16
Feature/odt service user

alex-delany:
Add ODT service-user Topic subscription and synapse RBAC


ChrisToppingKandC (approved):
Looks good, just need checks to pass.

PR #900, opened 2023-11-30
Mc bug case reference json fix

MikeCross1979:
changing document_case_reference.json so that case created date is string type


MariaValeroGonzalezKC (approved):
lgtm

PR #934, opened 2023-12-14
Feat/odw 886 create da rt api sql queries new

ChrisToppingKandC:
Initial work on DaRT queries.
dart_query - main sql query to be finalised
dart_test - test query to get extra fields and test the code etc
dart_cases_api - notebook to create a view and table in curated layer to be the final table to be queried by DaRT function app


MariaValeroGonzalezKC (approved):
lgtm

PR #953, opened 2024-01-04
Mc bug fix exam timetable

MikeCross1979:
Changing Examination Timetable source query to remove " chars as causing issues when storing as .CSV in raw


MariaValeroGonzalezKC (approved):
lgtm

PR #962, opened 2024-01-08
Mc bug fix service user case ref curated

MikeCross1979:
adding a left join to the NSIP data table in order to get the NSIP CaseReferences


MariaValeroGonzalezKC (approved):
lgtm

PR #988, opened 2024-01-17
Updating pipeline: 0_Horizon_Document_Folder - Source has had a column renamed

MikeCross1979:
Not much to this PR....

The document meta data source has changed at source, the column ValStr has been renamed to Author. Which means I need to update the mapping in this data copy task


MariaValeroGonzalezKC (approved):
lgtm

PR #990, opened 2024-01-17
Updating pipeline: 0_Raw_Horizon_NSIP_Relevant_Reps

MikeCross1979:
There is the need for an additional column to be added to Relevant Reps. New column is called AttachmentID and is needed in the ODT entity


MariaValeroGonzalezKC (approved):
lgtm

PR #1036, opened 2024-02-02
Updating notebook: relevant_representation

MikeCross1979:
Updated the case statement for representationFrom so that it follows the below logic from the business:

 Myself -> PERSON
An organisation -> ORGANISATION
Another Individual or Organisation -> AGENT

and "Members of the Public/Businesses" is actually one of the possible values for representationType


MariaValeroGonzalezKC (approved):
lgtm

PR #1051, opened 2024-02-14
Fix/curated validation none values

ChrisToppingKandC:
Added pandas code to notebooks to convert python None and NaT to empty strings to enable accurate validation.
Added to notebooks for each curated table.
Added master notebook to run all validations.


MariaValeroGonzalezKC (approved):
lgtm

PR #1178, opened 2024-04-29
Updating pipeline: pln_document_metadata_main

RobPalmer30:
Update pipeline to add call to service bus.

### DIFF
--- a/workspace/pipeline/pln_document_metadata_main.json
+++ b/workspace/pipeline/pln_document_metadata_main.json
@@ -18,6 +18,12 @@
 						"dependencyConditions": [
 							"Completed"
 						]
+					},
+					{
+						"activity": "pln_service_bus_nsip_document",
+						"dependencyConditions": [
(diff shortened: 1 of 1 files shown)


workspace/pipeline/pln_document_metadata_main.json
```diff
@@ -18,6 +18,12 @@
 						"dependencyConditions": [
 							"Completed"
 						]
+					},
+					{
+						"activity": "pln_service_bus_nsip_document",
+						"dependencyConditions": [
+							"Succeeded"
```
ChrisToppingKandC, line 25: Do we think succeeded is correct here? The other 2 pipelines have "completed" as the condition. Should all 3 be the same?

RobPalmer30: Updated so all 3 are on succeeded. On completion is causing failures to not be raised to parent pipelines

PR #1204, opened 2024-05-23
Updating notebook: casework_nsip_examination_timetable_dim_legacy

paul-dragan-spartaglobal:
**PR Template**

 1. Are there new source to raw datasets?
	
	 - [ ] If yes - has a trigger been attached at the appropriate frequency?
  	 - [ ] No

 2. Have any new tables been created in Standardised?

  	- [ ] No
   	- [ ] If yes:
		- [ ] orchestration.json has been updated and tested in Dev and has been / is about to be PRd into main
		- [ ] the new schema exists inside *odw-config/standardised-table-definitions* OR is about to be PRd into main
			- Make sure to run Platform Integrate and Platform Deploy to Dev at least to ensure the schema is deployed into Synapse Dev Live 
		- [ ] Is the raw-to-standardised python script scheduled to run for this dataset grouping?
 3. Have any new tables been created in Harmonised or Curated?

   	 - [ ] No
     	 - [ ] If yes
	 	- [ ]  *2-odw-standardised-to-harmonised/py_odw_harmonised_table_creation* OR 4-*odw-harmonised-to-curated/py_odw_curated_table_creation* are set up to run in the pipeline *pln_post_deployments* with the base parameter specifying the correct table
		- [ ] the new schema exists inside *odw-config/harmonised-table-definitions* / *odw-config/curated-table-definitions* OR is about to be PRd into main
			- Make sure to run Platform Integrate and Platform Deploy to Dev at least to ensure the schema is deployed into Synapse Dev Live 
 4. Have any tables changed AND/OR have any columns changed in any scripts?
We only care about new columns or columns that change type.
	- [ ] No
 	- [ ] If yes:
		- [ ] Please set py_change_table to run in the pipeline *pln_post_deployments*
		- [ ] Please create a script to backdate and fill in this new column in Test and Prod
			Only delete and recreate tables with caution!
 4. Have any scripts run in isolation in dev? Please look at the *"spark.autotune.trackingId"*
		- If these changes need to be reflected in Test and Prod, please add to the pipeline *post-			deployment/pln_post_deployment*
	- [ ] Yes - I have reflected this script in the *pln_post_deployments* pipeline
	- [ ] Yes - This script is part of a scheduled run and has been added to the appropriate end to end pipeline with a trigger at the correct frequency
	- [ ] No - This change does not need to take place in Test and Prod
	- [ ] No - No scripts have run

### DIFF
--- a/workspace/notebook/casework_nsip_examination_timetable_dim_legacy.json
+++ b/workspace/notebook/casework_nsip_examination_timetable_dim_legacy.json
@@ -21,7 +21,7 @@
 				"spark.dynamicAllocation.enabled": "false",
 				"spark.dynamicAllocation.minExecutors": "2",
 				"spark.dynamicAllocation.maxExecutors": "2",
-				"spark.autotune.trackingId": "7811cb26-af38-4602-baa3-e6b010575ab9"
+				"spark.autotune.trackingId": "5dc9c2c2-2a51-4ee3-887c-01eb5e500a57"
 			}
 		},
 		"metadata": {
@@ -107,6 +107,7 @@
 					"    T1.location                     AS Location,\r\n",
 					"    null                            AS EventLineItems,\r\n",
 					"    \"0\"                             AS Migrated,\r\n",
+					"    NULL                            AS Published,\r\n",
 					"    \"Horizon\"                       AS ODTSourceSystem,\r\n",
 					"    T2.SourceSystemID               AS SourceSystemID,\r\n",
 					"    to_timestamp(T1.expected_from)  AS IngestionDate,\r\n",
@@ -151,7 +152,7 @@
 					"    AND T1.expected_from = (SELECT MAX(expected_from) FROM odw_standardised_db.horizon_examination_timetable)\r\n",
 					";"
 				],
-				"execution_count": 5
+				"execution_count": 4
 			},
 			{
 				"cell_type": "markdown",
@@ -207,6 +208,7 @@
 					"    Location,\r\n",
 					"    EventLineItems,\r\n",
 					"    Migrated,\r\n",
+					"    Published,\r\n",
 					"    ODTSourceSystem,\r\n",
 					"    SourceSystemID,\r\n",
 					"    IngestionDate,\r\n",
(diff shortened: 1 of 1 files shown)


workspace/notebook/casework_nsip_examination_timetable_dim_legacy.json
```diff
@@ -107,6 +107,7 @@
 					"    T1.location                     AS Location,\r\n",
 					"    null                            AS EventLineItems,\r\n",
 					"    \"0\"                             AS Migrated,\r\n",
+					"    NULL                            AS Published,\r\n",
```
muhammadkc, line 110: Shouldn't this be T1.Published since we already have the data in standardised table?

ChrisToppingKandC (approved):

ChrisToppingKandC (changes requested):
Sorry, just seen Muhammad's comment. I'll leave this to you two to decide.

workspace/notebook/casework_nsip_examination_timetable_dim_legacy.json
RobPalmer30: One request, can we use descriptive names as aliases instead of T1. or a. please? It makes the code a lot easier to read as you don't have the cognitive overhead of remembering what the aliases represent.

paul-dragan-spartaglobal: While I do agree that in some case aliases might be better, in this particular case we are calling views and tables with the same/similar name and aliases will make things a bit more tricky. 
Also T1,T2,T3...etc was used across all harmonised tables when we initially build them, so changing them now would be quite a hassle.

RobPalmer30: We can address it going forwards as we touch code rather than retro fitting throughout. For SQL developers like myself it's poor practice to use aliases like this, we should be using something more obvious instead (e.g. T1 could be exam_timetable, etc).

RobPalmer30: Also the current join order is going to result in a poor execution plan due to the the join order. We should join Horizon timetable to casework first and then use a sub-select to inner join the subset of sourcesystem_fact we're interested in. That way we get the most optimum execution plan from the execution engine. 

The current join order will result in a busy plan and add overhead to the processing time.

RobPalmer30 (approved):

RobPalmer30 (approved):

ChrisToppingKandC (approved):

MariaValeroGonzalezKC (commented):
lgtm

PR #1243, opened 2024-06-06
Odw 1253 curated layer uniting data sources change

IsminiA:
run the curated notebook for appeals folder
updated the pipeline with the curated step for appeals folder


MariaValeroGonzalezKC (approved):
lgtm

PR #1259, opened 2024-06-19
Feat/historical data load filename

ChrisToppingKandC:
Added input file column to raw data historical function. When loading all historical data into a dataframe for a certain folder / entity this will add a column for the file name so you can see exactly what json file that particular row of data came from.


MariaValeroGonzalezKC (approved):
lgtm

PR #1371, opened 2024-08-22
Pd/fix/1380 nsip document curated

paul-dragan-spartaglobal:
nsip_document updated to match the required logic. 
Logic: all active cases must have active documents while the inactive cases must have inactive documents 
nsip_document


paul-dragan-spartaglobal:
https://pins-ds.atlassian.net/browse/ODW-1382

ChrisToppingKandC (approved):

PR #1417, opened 2024-09-13
Feat/s62a curated

paul-dragan-spartaglobal:
image
vanila curated similar to Gareth's view in minpins


ChrisToppingKandC:
@paul-dragan-spartaglobal can you create a unit test notebook for s62 and make sure it all passes? You can copy the appeal-document one and amend accordingly. Thanks.

paul-dragan-spartaglobal:
added s62a unit testing for data to go through all layers and match with multiple tables 
s62a vc

ChrisToppingKandC (approved):

PR #1430, opened 2024-09-19
Feat/1391 nsip project

IsminiA:
- Created new curated notebook : nsip_data_only_horizon
 - updated the py_unit_tests_functions with more tests for the new curated migatrion layer.
 - Updated py_unit_tests_nsip_project with more tests from the new migration curated layer.


workspace/notebook/py_unit_tests_functions.json
```diff
@@ -315,6 +315,128 @@
...
+							"deleting": false
+						}
+					}
+				},
+				"source": [
+					"## Define the harmonise table that only includes horizon data, and the latest record per primary id.\r\n",
+					"def harmonise_df_hzn(hrm_table: str, primary_key: str) -> DataFrame:\r\n",
+					"    df_harmonise_only_hzn: DataFrame = spark.sql(f\"\"\"\r\n",
+					"    select *\r\n",
+					"    from {hrm_db_name}.{hrm_table_final} tbl_1\r\n",
```
muhammadkc, line 337: in this function, the parameter `hrm_table` is not being used anywhere. And the two variables `hrm_db_name`, `hrm_table_final` that are being used don't exist. May be we can use the hrm_table instead?

PR #1437, opened 2024-09-23
disabled provider registration

ChrisToppingKandC:
Disabled provider registration. This is a temporary measure while we investigate more time to upgrade to version 4.

Pipeline succeeded - results here https://dev.azure.com/planninginspectorate/operational-data-warehouse/_build/results?buildId=93113&view=results


beejjacobs (approved):
I'd probably just apply this to the ODT and Horizon providers, but happy to approve this anyway.

PR #1458, opened 2024-10-02
Skip failing tests until we fix them

Syed-KC:
Currently we have these tests failing due to issues with their respective notebooks

- test_entraid.py
- test_function_gettimesheets.py
- test_nsip_s51_advice.py

There is already a bug raised for [test_entraid.py](https://pins-ds.atlassian.net/browse/THEODW-1389) and others will be added. As part of the bug tickets it is noted that the tests' should be turned back on as part of completing the ticket.


Syed-KC:
I'm closing this pull request because we have found the errors and there's no point skipping them now

PR #1488, opened 2024-10-15
Theodw 1396 update py unit tests appeals has

Syed-KC:
Added unit tests for py_unit_tests_appeals_has which has been renamed for consistency. Also updated a function and the corresponding call.


workspace/testing/tests/test_appeals_has.py
```diff
@@ -0,0 +1,66 @@
+import pytest
+import pipelineutils
+import constants
+import warnings
+
+def test_appeal_event_notebook(credential_name, azure_credential, synapse_endpoint: str):
```
ChrisToppingKandC, line 6: The function needs to be named correctly.

workspace/notebook/py_unit_tests_appeals_has.json
ChrisToppingKandC, line 1: Renaming of notebooks can often cause errors when publishing.

ChrisToppingKandC (changes requested):
Added a couple of comments.

workspace/notebook/py_unit_tests_appeals_has.json
Syed-KC: is there any way to do it without it causing issues?

workspace/testing/tests/test_appeals_has.py
Syed-KC: pushed a commit to fix this

workspace/notebook/py_unit_tests_appeals_has.json
ChrisToppingKandC: Yeah possibly directly i the json file rather than the UI but for now can you name it back to the original name so we can make progress and revisit the naming a bit later? Thanks.

Syed-KC: ok sure

Syed-KC: it's done

ChrisToppingKandC (approved):

PR #1489, opened 2024-10-16
creating vars for horizon and sub id in agent-plan, and also upgradin…

m-juckes-pins:
Infra upgrades
upgrading terraform to 1.9.6


workspace/trigger/tr_daily_7days_2100.json
ChrisToppingKandC, line 1: @m-juckes-pins can you remove this change please? Thanks. Otherwise OK.

ChrisToppingKandC (changes requested):
See comment.

workspace/trigger/tr_daily_7days_2100.json
m-juckes-pins: there was a git conflict so had to bring your change in. Not sure if you mean the extra line on the bottom? This will happen anyway as we should be using pre-commit across all projects as it is best practice

ChrisToppingKandC (approved):

PR #1499, opened 2024-10-24
Theodw 1405 update py unit tests service user notebook based on appeals document template and patterns

Syed-KC:
Added tests to service user
Updated py_unit_tests_functions and calls to it
Create pytest script to call service user notebook


workspace/notebook/py_unit_tests_service_user.json
```diff
@@ -598,7 +694,7 @@
 					"SELECT * \n",
 					"FROM\n",
 					"odw_curated_db.nsip_representation AS reps \n",
-					"INNER JOIN odw_curated_db.service_user AS suser\n",
+					"INNER JOIN odw_curated_db.appeal.service_user AS suser\n",
```
muhammadkc, line 697: @Syed-KC  is this supposed to be `odw_curated_db.appeal_service_user` instead?

Syed-KC: good catch! yeah it is supposed to be that, is it easy for you to change or shall I do a new commit?

muhammadkc: I'll update it no problem

muhammadkc (approved):

PR #1525, opened 2024-11-13
Theodw 1403 update py unit tests nsip s51 advice

Syed-KC:
Fixed tests on s51 advice that didn't make sense and added to the functions notebook
Updated pytest script

### DIFF
--- a/workspace/notebook/py_unit_tests_functions.json
+++ b/workspace/notebook/py_unit_tests_functions.json
@@ -20,7 +20,7 @@
 				"spark.dynamicAllocation.enabled": "false",
 				"spark.dynamicAllocation.minExecutors": "2",
 				"spark.dynamicAllocation.maxExecutors": "2",
-				"spark.autotune.trackingId": "f9abbf16-f0dd-4332-acda-ceeae6a9b838"
+				"spark.autotune.trackingId": "f50e3d23-95bb-46ee-ae1d-484b3ba55128"
 			}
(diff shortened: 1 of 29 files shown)


workspace/notebook/py_unit_tests_functions.json
```diff
@@ -776,6 +776,36 @@
...
+						"outputs_hidden": false
+					},
+					"nteract": {
+						"transient": {
+							"deleting": false
+						}
+					}
+				},
+				"source": [
+					"def test_no_project_from_horizon_dropped(std_db: str, curated_db_migration: str, horizon_table:str, curated_table_migration:str):\n",
```
muhammadkc, line 793: The objective of this test is to make sure no record from horizon is dropped. I think this could be made a bit more generic. i.e by making the primary key a parameter. I believe for s51 advice, the primary key is advicenodeid/adviceId instead of caseReference. 

Something like this?
`def test_no_record_from_horizon_dropped(std_db: str, curated_db_migration: str, horizon_table:str, curated_table_migration:str, horizon_pk: str, curated_pk: str)`

muhammadkc: And then we can pass `caseNodeId` as a horizon_pk from s51 advice unit test notebook, and `caseReference` from nsip_project unit test notebook

ChrisToppingKandC (approved):

PR #1554, opened 2024-12-04
Feat/theodw 885 listed buildings

muhammadkc:
Unit tests passed 
Screenshot 2024-12-09 at 12 19 52

### DIFF
--- a/workspace/pipeline/pln_publish_to_sb.json
+++ b/workspace/pipeline/pln_publish_to_sb.json
@@ -45,16 +45,13 @@
 		],
 		"parameters": {
 			"service_bus_name": {
-				"type": "string",
-				"defaultValue": "pins-sb-appeals-bo-dev"
+				"type": "string"
 			},
 			"messages": {
-				"type": "string",
-				"defaultValue": "This is a test message"
+				"type": "string"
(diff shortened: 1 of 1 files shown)


workspace/pipeline/pln_publish_to_sb.json
```diff
@@ -0,0 +1,65 @@
...
+			}
+		],
+		"parameters": {
+			"service_bus_name": {
+				"type": "string",
+				"defaultValue": "pins-sb-appeals-bo-dev"
+			},
+			"messages": {
+				"type": "string",
+				"defaultValue": "This is a test message"
```
RobPalmer30, line 53: Is this meant to be set to this still? Might be good to set it to something clearer that means it's a missing parameter

RobPalmer30 (commented):
Please see inline comment. Other than that looks good

workspace/pipeline/pln_publish_to_sb.json
muhammadkc: Good catch, shouldn't be there. Removing now.

RobPalmer30 (approved):

PR #1559, opened 2024-12-10
Feat/theodw 1392 suffix data

kathryn-thompson:


workspace/notebook/pins_inspectors.json
```diff
@@ -0,0 +1,538 @@
...
+					"    email,\r\n",
+					"    Migrated,\r\n",
+					"    ODTSourceSystem,\r\n",
+					"    IngestionDate,\r\n",
+					"    ValidTo,\r\n",
+					"    RowID,\r\n",
+					"    IsActive\r\n",
+					"    \r\n",
+					"FROM odw_harmonised_db.pins_inspectors\r\n",
+					"WHERE horizonId IN (SELECT horizonId FROM pins_inspectors_new WHERE horizonId IS NULL) AND IsActive = 'Y'; "
```
RobPalmer30, line 304: isn't this line always going to return everything? I don't think you need the where horizon Id is null clause

workspace/notebook/pins_inspectors.json
```diff
@@ -0,0 +1,538 @@
...
+				"source": [
+					"%%sql\n",
+					"\n",
+					"CREATE OR REPLACE TEMPORARY VIEW Loading_month\n",
+					"\n",
+					"    AS\n",
+					"\n",
+					"SELECT DISTINCT\n",
+					"    IngestionDate AS IngestionDate,\n",
+					"    to_timestamp(date_sub(IngestionDate,1)) AS ClosingDate,\n",
```
RobPalmer30, line 334: Is this necessarily the case? To me this reads as an assumption that stuff has run every day. Would be better to look this up based on the data we have in the ODW

workspace/notebook/pins_inspectors_curated.json
```diff
@@ -0,0 +1,170 @@
...
+					"collapsed": false
+				},
+				"source": [
+					"%%sql\n",
+					"\n",
+					"CREATE OR REPLACE VIEW odw_curated_db.vw_pins_inspectors_curated\n",
+					"\n",
+					"AS\n",
+					"\n",
+					"SELECT DISTINCT \n",
```
RobPalmer30, line 107: Please reformat this so the comma is at the start of the field name instead of the end - it makes debugging a lot easier in the future

RobPalmer30 (commented):
Please see associated comments

RobPalmer30 (approved):

PR #1566, opened 2024-12-11
DaRT function app

kathryn-thompson:


ChrisToppingKandC (changes requested):
Getting error Ambiguous column name 'caseReference'. Probably in the where clause.

ChrisToppingKandC (approved):

PR #1577, opened 2024-12-18
created appeal_event_final for harmonised

IsminiA:
uodated the data model for appeal event to have the harmonised final appeal event


alex-ghoth (approved):
Not sure why this markdown was left hanging. Looks fine.

PR #1584, opened 2025-01-08
Function App documentation updates

kathryn-thompson:


ChrisToppingKandC:
Thanks Kat. A couple of points to make. The Service User process is out of date and can happily stay in the archived file you've created. Then I think the DaRT and Timesheets descriptions could be copied into the `servicebus-functionapps` document. The `general-function-apps` is therefore not needed on its own as we can just have 1 main document. Or did you have something else in mind?

ChrisToppingKandC (changes requested):
Thanks Kat. A couple of points to make. The Service User process is out of date and can happily stay in the archived file you've created. Then I think the DaRT and Timesheets descriptions could be copied into the servicebus-functionapps document. The general-function-apps is therefore not needed on its own as we can just have 1 main document. Or did you have something else in mind?

ChrisToppingKandC (approved):

PR #1665, opened 2025-03-05
Update testing-pipelines.yml for Azure Pipelines

kathryn-thompson:


ChrisToppingKandC (changes requested):
@kathryn-thompson you might want to add an `always: true` as per this documentation - https://learn.microsoft.com/en-us/azure/devops/pipelines/process/scheduled-triggers?view=azure-devops&tabs=yaml#running-even-when-there-are-no-code-changes

PR #1711, opened 2025-03-26
feat(infrastructure): add "Storage Blob Data Owner" role to purview instance

beejjacobs:


HarrisonBoyleThomas:

PR #1763, opened 2025-04-04
fix(tooling): rolling back to old azure agent image

Medini-Aithal:
rolling back to old azure agent image with terraform version 1.9.6


stef-solirius (approved):

HarrisonBoyleThomas:

PR #1769, opened 2025-04-09
Feat/theodw 1649 make recommendations from pipeline audit

NHAL22:
archivePipelinesNotUsedByMasterPln.py script does the following:
- Extracts pipelines used by the master pipeline & recursively goes through these pipelines
- Uses this to find the pipelines not used by master
- Checks if the pipeline has been archived and if not, then archives it


stef-solirius (approved):
We can approve your PR, merge and publish it to DEV Live - it would then be safe to trigger the pln_master in DEV and ensure we did not break any link by archiving something we did not mean to.

PR #1776, opened 2025-04-10
Latest: Feat/Theodw 1649 make recommendations from notebook audit

NHAL22:
archiveNotebooksNotUsedByPipelinesFromMasterPln&Unreferenced.py script does the following:

- Extracts pipelines used by the master pipeline & recursively goes through these pipelines
- Uses this to find the pipelines not used by master and find pipelines that are not being run
- Extracts the notebooks for these set of pipelines under this conditions
- Using buildHierarchyOfPipelines.py script, the output generated a list of unreferenced notebooks. This was also used to archive any necessary notebooks and integrated within the script


HarrisonBoyleThomas:
I think the python script should be in slug case and not have a `&` character

HarrisonBoyleThomas:
Overall looks good

alex-ghoth (approved):
Greate to see all these notebooks being moved to a workspace archive/ folder

PR #1782, opened 2025-04-14
Theodw 1638 update cron schedule

HarrisonBoyleThomas:
Update the azure-pipelines cron schedule to run at 10am every day rather than 7pm


HarrisonBoyleThomas:
Wrong branch

PR #1783, opened 2025-04-14
Remove cron schedule and use PR triggers instead

HarrisonBoyleThomas:
The cron schedule approach won't work because the pipeline would not run against the new changes. Switching to using a PR trigger so the pipeline can run the most recent code changes

### DIFF
--- a/.github/workflows/only-test-merge-into-main.yaml
+++ b/.github/workflows/only-test-merge-into-main.yaml
@@ -4,7 +4,7 @@ on:
   pull_request:
     branches: 
       - main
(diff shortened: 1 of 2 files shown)


Fred83200 (dismissed):

alex-ghoth (dismissed):

Fred83200 (approved):

HarrisonBoyleThomas:
Not needed. https://github.com/Planning-Inspectorate/ODW-Service/pull/1785 is a better solution

PR #1784, opened 2025-04-14
Theodw 1638 dev cron job

HarrisonBoyleThomas:
Part of https://pins-ds.atlassian.net/browse/THEODW-1638

Update the schedule of the pipeline to prevent deadlock from moving from dev->testing branches.


HarrisonBoyleThomas:
Wrong target branch

PR #1787, opened 2025-04-14
Theodw 1638 update modification detection

HarrisonBoyleThomas:

Update git diff target branch to be dynamic, rather than being hardcoded to be `dev`.
Add parameter to allow the full deployment to be run if desired

### DIFF
--- a/pipelines/azure-pipelines.yaml
+++ b/pipelines/azure-pipelines.yaml
@@ -6,6 +6,10 @@ parameters:
 - name: runTests
   type: boolean
   default: true
+# Whether or not to deploy everything
+- name: fullDeployment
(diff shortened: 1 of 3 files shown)


HarrisonBoyleThomas:
Example of ticking all of the new boolean parameters: https://dev.azure.com/planninginspectorate/operational-data-warehouse/_build?definitionId=269&_a=summary

Example of leaving the new parameters as false: https://dev.azure.com/planninginspectorate/operational-data-warehouse/_build/results?buildId=111513&view=results

HarrisonBoyleThomas:
I cancelled the build which is why it is red. This change only affects the first stage of the build

PR #1788, opened 2025-04-15
Skip unit tests in prod env due to them currently containing some int…

HarrisonBoyleThomas:

Unit tests are usually safe to run in prod since they are lightweight and isolated, however all ODW tests are classed as unit tests, so we should skip these tests in the prod env until they have been broken down


HarrisonBoyleThomas:
The changes are working: https://dev.azure.com/planninginspectorate/operational-data-warehouse/_build/results?buildId=111522&view=results

I cancelled it because it takes a long time to run and the final step always fails due to failing tests - github just treats the whole run as a failure because of it

NHAL22 (approved):

PR #1792, opened 2025-04-15
Update env variable definition to use compile-time expression

HarrisonBoyleThomas:

The most recent auto-trigger from the testing branch seemed to deploy to the dev environment rather than the test env. This is due to the variable definition of `env` not allowing nested conditional assignment


Fred83200 (approved):

HarrisonBoyleThomas:
The new code worked, i just cancelled the build to save time

PR #1810, opened 2025-04-17
Theodw 1638 simplified process

HarrisonBoyleThomas:

The original CI/CD strategy had a design flaw caused by divergence between dev and main due to squashed commits. This resulted in the dev branch being deadlocked. Keeping 3 branches in sync would have been complicated and prone to mistakes.

This PR removes the `testing` branch from the process, and simplifies the process as a whole. Now the flow is

- Developer implements a new feature on branch X
- Developer raises a PR against the dev branch
- CICD is triggered, and runs against the dev (or build) env
- Code merged
- A PR is automatically raised from the dev branch to the main branch when new code is added
- At a set time every day, the dev branch is run against the test env
- Once merged, the dev branch is deleted then recreated automatically (based on main), and the main branch is run against prod


pipelines/steps/check-which-component-is-modified.yaml
```diff
@@ -16,8 +16,8 @@ steps:
     deployEverything=false
     echo "Source branch is '$sourceBranch'"
     if [[ $sourceBranch = dev ]]; then
-        targetBranch=testing
-    elif [[ $sourceBranch = testing ]]; then
+        echo "The source branch is dev, so forcing a full deployment"
+        deployEverything=true
```
HarrisonBoyleThomas, line 20: I added `deployEverything=true` here incase something else had been deployed to test, so that all new changes are deployed there. Theoretically this should be fast since only modified infra can ever be applied by Terraform.

i.e. the assumption is that dev/build envs will only be touched by this process, while test/main _could_ have a different state. The dev/build envs can be fully restored to a specific state by running the pipeline manually

stef-solirius (approved):

alex-ghoth (approved):

PR #1829, opened 2025-04-23
Feat/theodw 1753 synapse backup pool34 - replace pinssynspodw to pinssynspodw34

stef-solirius:
Replace pinssynspodw to pinssynspodw34 in notebooks


.github/workflows/dev-to-test.yml
```diff
@@ -0,0 +1,61 @@
+###
```
HarrisonBoyleThomas, line 1: This needs to be deleted - might be a dodgy rebase?

stef-solirius: Actioned

PR #1857, opened 2025-04-29
Update spark pool references

HarrisonBoyleThomas:
Update all references to pool `pinssynspodw34` to `pinssynspodw`

This uses a script defined as part of https://github.com/Planning-Inspectorate/ODW-Service/pull/1859


HarrisonBoyleThomas:
There are more additions than deletions. An example is in `workspace/notebook/py_temp_syn_employee_to_curated.json` . The addition is caused by adding a missing property that is specified by the spark pool - I think this should be ok

PR #1859, opened 2025-04-29
Theodw 1760 synapse spark pool update script

HarrisonBoyleThomas:
Ticket: https://pins-ds.atlassian.net/browse/THEODW-1760

Add script to automatically update spark pool references

PR #1863, opened 2025-04-30
Add script to update spark pool references

HarrisonBoyleThomas:
Ticket: https://pins-ds.atlassian.net/browse/THEODW-1760

Add script to automatically update spark pool references

This is a copy of https://github.com/Planning-Inspectorate/ODW-Service/pull/1859

### DIFF
--- a/pipelines/scripts/update_spark_pool_references.py
+++ b/pipelines/scripts/update_spark_pool_references.py
@@ -1,0 +1,256 @@
+from pipelines.scripts.synapse_artifact.synapse_spark_pool_util import SynapseSparkPoolUtil
+from typing import Dict, List, Any
+import argparse
+import json
+import os
+from copy import deepcopy
+import logging
+
+
+class SparkPoolReferenceUpdater():
+    """
+        Class for managing updating references to spark pools
+    """
+    def __init__(self, synapse_workspace_name: str, old_pool_name: str, new_pool_name: str):
+        """
+            Class for managing updating references to spark pools
+
+            :param synapse_workspace_name: The name of the synapse workspace to target
+            :param old_pool_name: The old spark pool we should replace reference to
+            :param new_pool_name: The spark pool we should use to replace references
+        """
+        self._synapse_workspace_name = synapse_workspace_name
+        self._spark_pool_cache = dict()
+        self._old_pool_name = old_pool_name
+        self._new_pool_name = new_pool_name
+    
+    def update_all_spark_pool_references(self):
+        """
+            Update all references to the old spark pool with the new spark pool
+        """
+        logging.info("Starting to replace spark pool references")
+        artifact_names = self.get_all_relevant_artifact_names()
+        logging.info("Performing replacement on the below files")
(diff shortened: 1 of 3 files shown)


pipelines/scripts/update_spark_pool_references.py
```diff
@@ -0,0 +1,256 @@
...
+            :return: The artifact with updated references
+        """
+        if artifact_path.startswith("workspace/notebook"):
+            is_notebook = True
+        elif artifact_path.startswith("workspace/pipeline"):
+            is_notebook = False
+        else:
+            raise ValueError(f"Artifact '{artifact_path}' is not a path to a Synapse notebook or pipeline")
+        with open(artifact_path, "r") as f:
+            artifact = json.load(f)
```
Fred83200, line 88: If a malformed or non-JSON file is present the script will crash here. Have you thought on adding a try/except around it ? 
Something like this : 
try:
    artifact = json.load(f)
except json.JSONDecodeError as e:
    logging.error(f"Failed to parse JSON in {artifact_path}: {e}")
    return {}

pipelines/scripts/update_spark_pool_references.py
```diff
@@ -0,0 +1,256 @@
...
+        logging.info(json.dumps(artifact_names, indent=4))
+        logging.info("\nBegin replacement")
+        cleaned_artifacts = {
+            artifact_name: self.replace_spark_pool_references_in_artifact(artifact_name)
+            for artifact_name in artifact_names
+        }
+        logging.info("\nSaving changes\n")
+        for artifact_path, artifact_json in cleaned_artifacts.items():
+            logging.info(f"Overwriting artifact '{artifact_path}'")
+            with open(artifact_path, "w") as f:
```
Fred83200, line 43: So you overwrite the updated artifacts here - if something goes wrong mid-script it's risky.
We don't want to add an option to save original files in case ? in a backup folder ? or append with .bak just as safety ?

pipelines/scripts/update_spark_pool_references.py
HarrisonBoyleThomas: I'll make a change to improve the logging

pipelines/scripts/update_spark_pool_references.py
HarrisonBoyleThomas: This is intentional, the purpose of this script is mainly to help with the roll out of spark pool updates. An example of it being applied is https://github.com/Planning-Inspectorate/ODW-Service/pull/1862 and

pipelines/scripts/update_spark_pool_references.py
HarrisonBoyleThomas: I've made the change

pipelines/scripts/update_spark_pool_references.py
HarrisonBoyleThomas: If something is wrong with the json, it'll be caught during cicd when deploying synapse

Fred83200 (approved):

PR #1864, opened 2025-04-30
Hot fix/theodw 1751 valid date issue casevaliddate field in has entity mismatch between cbos and odw

roytalari:
Modified two notebooks to address the prod bug THEODW-1751 utc to daylight for MiPINS in the curated layer. 
1) appeal_s78_curated_mipins
2) appeal_has_curated_mipins

### DIFF
--- a/workspace/notebook/appeal_s78_curated_mipins.json
+++ b/workspace/notebook/appeal_s78_curated_mipins.json
@@ -20,7 +20,7 @@
 				"spark.dynamicAllocation.enabled": "false",
 				"spark.dynamicAllocation.minExecutors": "2",
 				"spark.dynamicAllocation.maxExecutors": "2",
-				"spark.autotune.trackingId": "6ad8415f-f1ab-4fec-b176-0f54d93d1d75"
+				"spark.autotune.trackingId": "92c564a2-935b-4886-8d17-877cb948a4a0"
 			}
 		},
 		"metadata": {
@@ -51,6 +51,16 @@
 			"sessionKeepAliveTimeout": 30
 		},
 		"cells": [
+			{
+				"cell_type": "code",
+				"source": [
+					"from pyspark.sql import SparkSession\n",
+					"from notebookutils import mssparkutils\n",
+					"from datetime import datetime, date\n",
+					"from pyspark.sql.functions import *"
+				],
+				"execution_count": 30
+			},
 			{
 				"cell_type": "code",
 				"metadata": {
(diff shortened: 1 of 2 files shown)


workspace/notebook/appeal_s78_curated_mipins.json
```diff
@@ -206,24 +216,68 @@
 					"FROM odw_harmonised_db.sb_appeal_s78 AS AH\n",
 					"where AH.lpaCode NOT in ('Q9999', 'Q1111')"
 				],
-				"execution_count": 28
+				"execution_count": 39
+			},
+			{
+				"cell_type": "code",
+				"source": [
+					"spark = SparkSession.builder.getOrCreate()\n",
```
HarrisonBoyleThomas, line 224: You shouldn't need to do this, `spark` is usually a notebook session-level object

workspace/notebook/appeal_s78_curated_mipins.json
```diff
@@ -206,24 +216,68 @@
...
+				],
+				"execution_count": 40
+			},
+			{
+				"cell_type": "code",
+				"metadata": {
+					"collapsed": false
+				},
+				"source": [
+					"# prepare dataframe convert UTC to Day light savings for all timestamp fields\n",
```
HarrisonBoyleThomas, line 235: Thinking, could it be simpler to define this as a loop instead, since the arguments to `from_utc_timestamp` are the same for each `withColumn` call ?

HarrisonBoyleThomas: Something like

```
col_list = [
    "caseSubmittedDate",
    "caseCreatedDate",
    ...
]
final_view_df = view_df
for column in col_list:
    final_view_df = final_view_df.withColumn(column, from_utc_timestamp(column, "Europe/London")
```

workspace/notebook/appeal_s78_curated_mipins.json
```diff
@@ -51,6 +51,16 @@
 			"sessionKeepAliveTimeout": 30
 		},
 		"cells": [
+			{
+				"cell_type": "code",
+				"source": [
+					"from pyspark.sql import SparkSession\n",
```
HarrisonBoyleThomas, line 57: This import can be removed based on my previous comment

workspace/notebook/appeal_s78_curated_mipins.json
```diff
@@ -51,6 +51,16 @@
...
+					"from notebookutils import mssparkutils\n",
+					"from datetime import datetime, date\n",
+					"from pyspark.sql.functions import col, from_utc_timestamp"
```
HarrisonBoyleThomas, line 60: I would avoid importing `col` like this, since `col` is commonly used as a variable name (not in this notebook but in general)

workspace/notebook/appeals_has_curated_mipins.json
```diff
@@ -51,6 +51,16 @@
 			"sessionKeepAliveTimeout": 30
 		},
 		"cells": [
+			{
+				"cell_type": "code",
+				"source": [
+					"from pyspark.sql import SparkSession\n",
```
HarrisonBoyleThomas, line 57: Same comments for this file too

workspace/notebook/appeal_s78_curated_mipins.json
roytalari: yes, that's right, corrected.

PR #1867, opened 2025-05-01
Define spark config as a constant. Update spark config references to …

HarrisonBoyleThomas:
Ticket: https://pins-ds.atlassian.net/browse/THEODW-1778

Define spark config as a constant in the tests, to reduce code duplication and simplify the process of updating the spark config (e.g if we need to delete a deprecated spark pool)


Fred83200 (approved):

alex-ghoth (approved):
Great, these notebooks needed to be switched over to the SPARK_POOL_CONFIG.

HarrisonBoyleThomas:
The tests are failing, however it is just assertion errors (which were failing before this PR)

PR #1878, opened 2025-05-07
Update service connection references

HarrisonBoyleThomas:
Ticket: https://pins-ds.atlassian.net/browse/THEODW-1639

Some of the ADO pipelines use a redundant service connection. We plan to avoid this mistake with the new build environment, so this PR updates all references to use the main service connection throughout all ADO processes


Fred83200 (dismissed):

Fred83200 (approved):

HarrisonBoyleThomas:
Not needed - it turns out the service connection is required for now because it uses credential kind required for some ADO tasks

PR #1894, opened 2025-05-15
Theodw 1804 approve mp es

HarrisonBoyleThomas:
Ticket: https://pins-ds.atlassian.net/browse/THEODW-1804

Add a script to auto-approve private endpoint connections. Add new PE approval script to the end of the main pipeline

Sample run against the dev env: https://dev.azure.com/planninginspectorate/operational-data-warehouse/_build/results?buildId=114550&view=logs&j=c8c38e96-1466-5c13-c06f-f8728efccd7e&t=bbee36d8-4858-53f9-f679-f15a24bf3b9c

### DIFF
--- a/pipelines/scripts/private_endpoint/private_endpoint_manager.py
+++ b/pipelines/scripts/private_endpoint/private_endpoint_manager.py
@@ -1,5 +1,4 @@
-from azure.identity import AzureCliCredential
-from typing import List, Dict, Any, Type, Iterable
+from typing import List, Dict, Any, Iterable
 from abc import ABC, abstractmethod
 from pipelines.scripts.util import Util
 import logging
@@ -7,7 +6,7 @@ import json
 
 class PrivateEndpointManager(ABC):
 
-    def get(self, private_endpoint_name: str, resource_group_name: str, resource_name: str) -> Dict[str, Any]:
+    def get(self, private_endpoint_id: str, resource_group_name: str, resource_name: str) -> Dict[str, Any]:
         """
             Return the details of the given private endpoint
         """
@@ -15,7 +14,7 @@ class PrivateEndpointManager(ABC):
             "az network private-endpoint-connection show"
         )
         command_args = {
-            "--name": private_endpoint_name,
+            "--id": private_endpoint_id,
             "--resource-group": resource_group_name,
             "--resource-name": resource_name,
             "--type": self.get_resource_type()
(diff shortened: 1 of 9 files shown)


pipelines/scripts/private_endpoint/private_endpoint_manager.py
```diff
@@ -0,0 +1,110 @@
...
+class PrivateEndpointManager(ABC):
+
+    def get(self, private_endpoint_id: str) -> Dict[str, Any]:
+        """
+            Return the details of the given private endpoint
+
+            :param private_endpoint_id: The private endpoint to fetch the details for
+        """
+        command_to_run = (
+            "az network private-endpoint-connection show"
```
Fred83200, line 17: These implicitly return JSON, but to ensure consistency across all environments maybe you can add (--output json) to every CLI command ?

pipelines/scripts/util.py
```diff
@@ -11,9 +14,59 @@ def run_az_cli_command(cls, args: List[str]):
         """
             Run an az cli command. Raises a `RuntimeException` if something goes wrong
         """
-        print(f"Running command: '{' '.join(args)}'")
+        logging.info(f"Running command {' '.join(args)}")
         try:
             return subprocess.check_output(args)
         except subprocess.CalledProcessError as e:
             exception = e
```
Fred83200, line 21: Might be helpful to maybe show the command output/error message explicitly or log the full AZ CLI command that failed to debug ? Might be easier ? Just a thought

pipelines/steps/approve-odw-private-endpoints.yaml
```diff
@@ -0,0 +1,17 @@
...
+  env: ''
+  pythonVersion: ''
+
+# Note: Although this can approve all private endpoints, this is only really needed for Managed private endpoints created by the Synapse deployment
+steps:
+  - script: |
+      set -x
+      python${{ parameters.pythonVersion }} --version
+      export PYTHONPATH="${PYTHONPATH}:/./"
+      python${{ parameters.pythonVersion }} $(System.DefaultWorkingDirectory)/pipelines/scripts/approve_private_endpoints.py --env "${{ parameters.env }}"
```
Fred83200, line 11: Does this vary ? Maybe using a variable or a central script root might be better ? 
Something like this : 
scriptRoot: $(Build.SourcesDirectory)/pipelines/scripts

pipelines/scripts/util.py
HarrisonBoyleThomas: I want to explicitly show the command just for debugging purposes in the future, it's helpful to see what command was run in ADO if troubleshooting something going wrong

pipelines/scripts/private_endpoint/private_endpoint_manager.py
HarrisonBoyleThomas: I'll add this in

pipelines/steps/approve-odw-private-endpoints.yaml
HarrisonBoyleThomas: We can't define a variable at this level. The only way we could do it is by adding a parameter, but this could then result in misuse of the script (by passing the wrong path)

pipelines/scripts/util.py
Fred83200: Completely agree you're already doing part of this with this command. 

I suggest maybe this improvement to it : 

try:
    return subprocess.check_output(args, stderr=subprocess.STDOUT)
except subprocess.CalledProcessError as e:
    command_str = ' '.join(args)
    raise RuntimeError(f"Error running command: {command_str}\nOutput:\n{e.output.decode('utf-8')}")
    
    
    So you get the command that was run, output produced and why it failed

HarrisonBoyleThomas: I'd prefer to use the current approach, i want to always log the command just for visibility since debugging ADO can be tricky sometimes, the full trace can be helpful incase there is a transient error/mistake in the command

Fred83200 (approved):

PR #1895, opened 2025-05-15
Theodw 1811 clean notebooks

HarrisonBoyleThomas:
Ticket: https://pins-ds.atlassian.net/browse/THEODW-1811

Add script to validate Synapse notebooks by ensuring the `execution_count` cell is null. Add script to clean invalid notebooks flagged by the validation script. Clean all synapse notebooks using the cleaner script. Add github action to validate Synapse notebooks for new PRs

### DIFF
--- a/.github/workflows/code-validation.yaml
+++ b/.github/workflows/code-validation.yaml
@@ -16,4 +16,5 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Validate Synapse notebooks
       run: |
+        export PYTHONPATH="${PYTHONPATH}:/./"
         python3 pipelines/scripts/check_notebooks_valid.py
--- a/.github/workflows/code-validation.yaml
+++ b/.github/workflows/code-validation.yaml
@@ -1,6 +1,6 @@
 name: Code Validation
 
-on: [push]
+on: [pull_request]
 
 jobs:
   build:
--- a/.github/workflows/code-validation.yaml
+++ b/.github/workflows/code-validation.yaml
@@ -1,0 +1,19 @@
+name: Code Validation
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.10"]
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Validate Synapse notebooks
+      run: |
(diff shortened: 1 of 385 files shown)


pipelines/scripts/check_notebooks_valid.py
```diff
@@ -0,0 +1,60 @@
...
+
+        :param notebook: The notebook json to analyse
+
+        :return: `True` if the execution count is null, `False` otherwise
+    """
+    return all(
+        [
+            x
+            for x in notebook["properties"]["cells"]
+            if x.get("execution_count", None) is None
```
Fred83200, line 32: This cell will always return true if there's at least one cell with execution_count == none but it doesn't check all cells 

shouldn't it be like this : 
return all(
    cell.get("execution_count") is None
    for cell in notebook["properties"]["cells"]
)

So you check that every cell has execution_count == None

.github/workflows/code-validation.yaml
```diff
@@ -0,0 +1,20 @@
+name: Code Validation
+
+on: [push]
```
Fred83200, line 3: Is this set for a specific branch ? Cause for now this will trigger one every branch and every push? Do you want to narrow it down or is it intended ?

pipelines/scripts/clean_notebooks.py
```diff
@@ -0,0 +1,49 @@
...
+        Clean all of the given notebooks
+
+        :param notebooks_to_clean: The list of notebook paths to clean
+    """
+    for notebook_path in notebooks_to_clean:
+        logging.info(f"Cleaning notebook '{notebook_path}'")
+        notebook = json.load(open(notebook_path))
+        cleaned_notebook = clean_notebook(notebook)
+        if cleaned_notebook:
+            logging.info("    Some properties of this notebook have been cleaned")
```
Fred83200, line 43: Here can we show what was cleaned ? or how many cells where affected ?

.github/workflows/code-validation.yaml
HarrisonBoyleThomas: Could do tbf, lemme change it so it's just for PRs

pipelines/scripts/clean_notebooks.py
HarrisonBoyleThomas: We could show the json diff, but we'd need to import a new package to do it which i want to avoid. Imo there's not much value in the json diff, since the diff can be easily shown in vscode/gitub. I could easily extend the code to output the number of modified cells, so i'll add this in

Fred83200 (dismissed):

Fred83200 (approved):

PR #1921, opened 2025-05-29
Feat/theodw 1776 singular listed building

stef-solirius:
Data-model for listed-building shows that $id and filenames for schemas do not match. 

The listed-building notebook should has been updated to reference the singular entity.

Set entity_name to listed-building (singular)


roytalari (approved):
looks good

PR #1922, opened 2025-05-29
Feat/theodw 1759 rev implement delta backup logic across the odw

roytalari:
A pipeline to backup both delta and non-delta files and directories in standardised layer. 
This is a standalone pipeline to monitor backup process runs as expected. 
There is parameter to target container, which is temporary for testing. 
A trigger is added to run the pipeline 9am daily.


workspace/linkedService/ls_delta_backup.json
stef-solirius, line 1: Looks good, could we add some descriptive text at the beginning that explains what the three notebooks do:
- delta_back_odw
- delta_back_odw_validate
- py_delta_backup_table

roytalari: thanks, I have added some descriptive text to each notebook.

stef-solirius (approved):

PR #1963, opened 2025-06-12
Update unit tests

HarrisonBoyleThomas:
Ticket: https://pins-ds.atlassian.net/browse/THEODW-1881

Add two more sample unit tests that run Synapse notebooks

This is a copy of https://github.com/Planning-Inspectorate/ODW-Service/pull/1956 with a clean history

### DIFF
--- a/tests/util/notebook_run_test_case.py
+++ b/tests/util/notebook_run_test_case.py
@@ -24,14 +24,20 @@ class NotebookRunTestCase(SynapseTestCase):
         """
         logging.info(f"RUNNING notebook {notebook_name}...\n")
         notebook_run_id = str(uuid.uuid4())
-        run_notebook_url = f'{self.SYNAPSE_ENDPOINT}/notebooks/runs/{notebook_run_id}?api-version=2022-03-01-preview'
+        run_notebook_url = f"{self.SYNAPSE_ENDPOINT}/notebooks/runs/{notebook_run_id}?api-version=2022-03-01-preview"
         headers = {"Authorization": f"Bearer {self.SYNAPSE_ACCESS_TOKEN}", "Content-Type": "application/json"}
-        response = requests.put(run_notebook_url, headers=headers,data=json.dumps(notebook_parameters))
+        run_parameters = {
+            "notebook": notebook_name,
+            "parameters": notebook_parameters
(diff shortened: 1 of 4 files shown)


tests/util/notebook_run_test_case.py
```diff
@@ -24,14 +24,20 @@ def _trigger_notebook(self, notebook_name: str, notebook_parameters: Dict[str, A
...
+        if notebook_parameters:
+            run_parameters["parameters"] = notebook_parameters
+        response: requests.Response = requests.put(run_notebook_url, headers=headers, data=json.dumps(run_parameters))
         if response.status_code >= 200 and response.status_code < 400 : 
             return notebook_run_id
-        raise NotebookRunException(f"Failed to run notebook {notebook_name}. RunID: '{notebook_run_id}'. Error is {response.status_code}...\n")
+        raise NotebookRunException(f"Failed to run notebook {notebook_name}. RunID: '{notebook_run_id}'. Error is {response.text}...\n")
 
-    def _wait_for_notebook_run(self, notebook_run_id: str, poll_interval = 15, max_wait_time_minutes: int = 10) -> str:
+    def _wait_for_notebook_run(self, notebook_run_id: str, poll_interval = 15, max_wait_time_minutes: int = 10) -> Dict[str, Any]:
```
Fred83200, line 40: Your wait here could hang longer if the notebook is stuck and not returning Succeeded or a state back, maybe you can log some results or status ? (running, queued ...) to trace it better ?

Fred83200 (approved):

HarrisonBoyleThomas:
> Are these unit_test ? They look more lie integration test as they run actual notebooks via the RestAPI right ?

These are unit tests for Synapse notebooks - they test the most basic notebooks that have no dependencies. The only way we can run them is through the REST API

tests/util/notebook_run_test_case.py
HarrisonBoyleThomas: Thinking about it, long-term notebooks will be queued for ages, which will be a bottleneck. I might change it so they're all grouped together in a single notebook so only 1 runs, instead of running loads in parallel

Fred83200 (approved):

PR #1965, opened 2025-06-13
chore(infrastructure): upgrade to terraform 1.12.2

beejjacobs:
https://pins-ds.atlassian.net/browse/DEV-468
https://pins-ds.atlassian.net/browse/DEV-397


pipelines/devops-agent-deploy.yaml
```diff
@@ -51,78 +51,7 @@ pool:
   vmImage: ubuntu-22.04  # Ubuntu-latest by default does not have Terraform installed
 
 stages:
-  - stage: Plan
```
HarrisonBoyleThomas, line 54: We're planning to eventually migrate this to its own repo, but i think the infra deployment is needed since the main infra deployment doesn't deploy this pat of the infra

HarrisonBoyleThomas (approved):

PR #2016, opened 2025-07-04
Theodw 1749 unarchive all artifacts

HarrisonBoyleThomas:
Ticket: https://pins-ds.atlassian.net/browse/THEODW-1749

Unarchive all artifacts, since some have been archived that shouldn't have been


HarrisonBoyleThomas:
I have only targeted anything with `archive` in its name. I have also seen there are a few with `unused` in their name - however the new archival process will treat this as an unarchived files so it will still work as expected

PR #2020, opened 2025-07-04
Theodw 1749 auto archive artifacts

HarrisonBoyleThomas:
Ticket: https://pins-ds.atlassian.net/browse/THEODW-1749

Add a github action to auto archive/delete unused synapse artifacts, which is raised a separate PR

Update `SynapseArtifactUtil` to not initialise the connection to Azure when the module loads - instead it is only loaded if specifically requested, to simplify the usage of the util classes on github without needing to setup more identities, since they're not really needed for this new archival process

This PR does not actually archive/delete anything - this is scheduled to run at 9am on monday.

### DIFF
--- a/.github/workflows/synapse-artifact-archiver.yaml
+++ b/.github/workflows/synapse-artifact-archiver.yaml
@@ -34,7 +34,7 @@ jobs:
 
             if [[ `git status --porcelain` ]]; then
                 git add -A
-                git commit -m "Auto archive Synapse artifacts"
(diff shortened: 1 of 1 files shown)


HarrisonBoyleThomas:
Sample PR that was auto raised by https://github.com/Planning-Inspectorate/ODW-Service/actions/runs/16074172687/job/45365342349

https://github.com/Planning-Inspectorate/ODW-Service/pull/2019

stef-solirius (approved):

PR #2065, opened 2025-07-28
Feat/theodw 1944 pilot migrate logging database to artifact based external tables

roytalari:
- Creating test tables as External tables based on artifacts
- Added database_util to synape_artifact_factory 
- This is to validate and test if the database and tables are deployable when migrating to Test environment.

### DIFF
--- a/workspace/database/logging_db/table/delta_backup_odw_v2.json
+++ b/workspace/database/logging_db/table/delta_backup_odw_v2.json
@@ -5,7 +5,7 @@
       "Name": "delta_backup_odw_v2",
       "EntityType": "TABLE",
       "Namespace": {
-        "DatabaseName": "logging"
+        "DatabaseName": "logging_db"
       },
       "Description": "",
       "TableType": "EXTERNAL",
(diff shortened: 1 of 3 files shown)


roytalari:
Testing creation of logging database/ test external tables

roytalari:
> I only think the new util classes should be merged

pipelines/scripts/synapse_artifact/synapse_database_util.py
```diff
@@ -0,0 +1,115 @@
...
+                break
+            data = next_response.json()
+            all_tables.extend(data.get("items", []) or data.get("value", []))
+
+        return all_tables
+    
+    def get_table(self, database_name: str, table_name: str) -> Dict[str, Any]:
+        url = f"{self.synapse_endpoint}/databases/{database_name}/tables/{table_name}?api-version=2021-04-01"
+        response = self._web_request(url)
+        return response.json()
```
HarrisonBoyleThomas: These probably need to be moved to a new util class, since tables seem to be a different artifact kind

PR #2068, opened 2025-07-29
Updating pipeline: AIEData

Fred83200:
[https://pins-ds.atlassian.net/browse/THEODW-2000](url)

This PR introduces performance and reliability improvements to the AIEData pipeline by addressing delays and concurrency issues during file ingestion.

Changes Implemented:

-- Increased Copy Activity Timeout
Updated the Copy File activity timeout from 10 minutes to 20 minutes to accommodate possible delays from the integration runtime and queueing

-- Replaced Random Delay with Fixed Wait
Replaced the previous dynamic delay (@rand(1, 10) seconds) in the Wait activity with a fixed delay of 30 seconds to ensure consistent staggering of copy operations

-- Introduced Controlled Concurrency via Batching
Enabled batching in the For Each AIE Document loop with a Batch count of 2 to reduce pressure on the Integration Runtime and avoid overloading the pipeline with parallel copy activities

### DIFF
--- a/workspace/pipeline/AIEData.json
+++ b/workspace/pipeline/AIEData.json
@@ -65,6 +65,7 @@
 						"type": "Expression"
 					},
 					"isSequential": false,
+					"batchCount": 2,
 					"activities": [
 						{
 							"name": "Copy File",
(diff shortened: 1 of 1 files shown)


roytalari (approved):
looks good

PR #2090, opened 2025-08-19
Feat/theodw 2197 cicd concurrency

HarrisonBoyleThomas:
Ticket: https://pins-ds.atlassian.net/browse/THEODW-2197

Update the Synapse deployment jobs to use the ADO environments so that stage-level exclusive locks can be applied (i.e only 1 deployment can run in the env at a time, but concurrent deployments to different environments is possible)

### DIFF
--- a/pipelines/jobs/deploy-synapse.yaml
+++ b/pipelines/jobs/deploy-synapse.yaml
@@ -5,12 +5,13 @@ parameters:
   activateTriggers: false
   primaryWorkspace: true
   fullDeployment: false
+  adoEnvName: ''
 
 ##
 # Deploy Azure Synapse
 ##
 jobs:
-- job: DeploySynapseJob
+- deployment: DeploySynapseJob
   pool: ${{ parameters.agentPool }}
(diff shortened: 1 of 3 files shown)


pipelines/jobs/deploy-synapse.yaml
```diff
@@ -27,95 +28,103 @@ jobs:
...
-    clean: true
-  - template: ${{variables['System.DefaultWorkingDirectory']}}/pipelines/steps/azure-login.yaml@odw-common
-  - template: ${{variables['System.DefaultWorkingDirectory']}}/pipelines/steps/build-and-deploy-odw-package.yaml
-    parameters:
-      armServiceConnectionName: ${{ parameters.armServiceConnectionName }}
-      env: ${{ lower(parameters.env) }}
+  environment: ${{ parameters.adoEnvName }}
+  strategy:
+    runOnce:
+      deploy:
```
HarrisonBoyleThomas, line 122: These lines were added, everything else was indented

stef-solirius (approved):

PR #2129, opened 2025-09-22
Improve error handling around test case

HarrisonBoyleThomas:
Improve error handling for test case in rare case where there is no exit value from the notebook


HarrisonBoyleThomas:
Not needed, this is replaced by https://github.com/Planning-Inspectorate/odw-synapse-workspace/pull/2131

PR #2131, opened 2025-09-22
Theodw 2298 fix smoke test

HarrisonBoyleThomas:
Ticket: https://pins-ds.atlassian.net/browse/THEODW-2298

Improve error message to also include the stack trace of the failing test from the notebook

### DIFF
--- a/tests/smoke_test/test_smoke_synapse_connectivity.py
+++ b/tests/smoke_test/test_smoke_synapse_connectivity.py
@@ -1,5 +1,5 @@
 from tests.util.notebook_run_test_case import NotebookRunTestCase
-import json
+import ast
 
 
 class TestSmokeSynapseConnectivity(NotebookRunTestCase):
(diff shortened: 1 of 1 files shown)


workspace/notebook/test_smoke_py_connectivity.json
```diff
@@ -116,9 +116,8 @@
 					"        print(exception)\n",
 					"        print()\n",
 					"    \n",
-					"    failing_tests = list(exception_map.keys())\n",
-					"    mssparkutils.notebook.exit(failing_tests)\n",
-					"mssparkutils.notebook.exit([])"
+					"    mssparkutils.notebook.exit(exception_map)\n",
+					"mssparkutils.notebook.exit(dict())"
```
Fred83200, line 120: You're having two exits here ? only the last one run so you will always exit with an empty dict and mask the real failure

tests/smoke_test/test_smoke_synapse_connectivity.py
```diff
@@ -17,5 +17,9 @@ def test_synapse_connectivity(self):
 
         notebook_run_result = self.run_notebook(notebook_name, notebook_parameters)
         exit_value = notebook_run_result["result"]["exitValue"]
-        failing_tests = json.loads(exit_value)
-        assert not failing_tests, f"The following tests failed in the test_smoke_py_connectivity notebook : {json.dumps(failing_tests, indent=4)}"
+        if exit_value:
+            exit_value_cleaned = ast.literal_eval(exit_value)
```
Fred83200, line 21: Using literal_eval it can come back as string ? Isn't it better to use JSON end to end ? do a json.dumps and the test do a json.loads(exit_value) ? so you don't have surprises with the eval ?

HarrisonBoyleThomas: Good idea, i'll add this in

PR #2136, opened 2025-09-26
Feat/theodw 2306

Fred83200:
[https://pins-ds.atlassian.net/browse/THEODW-2306](url)

Added joins to the following Horizon views:
- Horizon_vw_curr_TypeOfLevel
- Horizon_vw_BIS_SpecialistCaseDates
- Horizon_vw_BIS_PlanningAppStrings
- Horizon_vw_BIS_PlanningAppDates
- Horizon_vw_BIS_LeadCase
- Horizon_vw_BIS_CaseStrings
- Horizon_vw_BIS_CaseInfo
- Horizon_vw_BIS_CaseDates
- Horizon_vw_BIS_AppealsAdditionalData
- Horizon_TypeOfProcedure

- Introduced deduplication logic (ROW_NUMBER() over case keys) to prevent row fan-out from the new joins and ensure only the latest row per case is kept
- Mapped and exposed all new fields from the spreadsheet (allocation, planning, decision, validation, appeals, and procedure fields)
- Dropped the helper column level_code used for joins so it does not appear in the final schema.
- Collapse multiple casespecialism values per case into a comma-separated string.
- Align the table name to follow the appeal-s78 documentation convention, producing the unified table:
odw_standardised_db.horizon_appeal_s78

Ensured the final output schema is the union of:
All legacy columns from the old join & All new mapped columns

TEST : 
Compared old vs. new outputs to confirm:

- Row counts and distinct caseuniqueid values match
- No missing or duplicated rows introduced after deduplication
- All legacy columns are still present
- All new mapped columns are included
- No unexpected extra columns remain.

The Notebook now produces the same data as before while adding the new Horizon-derived fields exactly as per the mapping spreadsheet. This ensures downstream consumers see no regressions, with the benefit of additional required attributes.


workspace/notebook/py_hzn_standardized_s78_tables.json
stef-solirius, line 1: Worth removing usused imports if we don't need them (IntegerType, ArrayType, StructType, StructField, Row) and the wildcard import (    # from pyspark.sql.functions import *)

workspace/notebook/py_hzn_standardized_s78_tables.json
```diff
@@ -184,25 +184,384 @@
...
+					"  LEFT JOIN af      ON h.caseuniqueid = af.appealrefnumber\r\n",
+					"),\r\n",
+					"\r\n",
+					"-- Collapse each Horizon red source to 1 row per key to avoid having some fan-out\r\n",
+					"pas_1row AS (\r\n",
+					"  SELECT * FROM (\r\n",
+					"    SELECT p.*,\r\n",
+					"           ROW_NUMBER() OVER (\r\n",
+					"             PARTITION BY p.casenodeid\r\n",
+					"             ORDER BY p.expected_from DESC, p.modified_datetime DESC, p.file_id DESC\r\n",
```
stef-solirius, line 278: Have we checked if expected_from and modified_datetime are always populated? If they can be null or identical across many rows, we might need to extend the ORDER BY with extra columns like file_id or ingestion_ts.

workspace/notebook/py_hzn_standardized_s78_tables.json
Fred83200: Done

workspace/notebook/py_hzn_standardized_s78_tables.json
```diff
@@ -193,16 +208,239 @@
...
+					"    aad_old.procedurelpa,\r\n",
+					"    aad_old.inspectorneedtoentersite,\r\n",
+					"    aad_old.sitegridreferenceeasting,\r\n",
+					"    aad_old.sitegridreferencenorthing,\r\n",
+					"    aad_old.sitewithinsssi,\r\n",
+					"    af.importantinformation,\r\n",
+					"    aad_old.level AS level_code\r\n",
+					"  FROM h\r\n",
+					"  LEFT JOIN cs_agg  ON h.caseuniqueid = cs_agg.casereference\r\n",
+					"  LEFT JOIN cd      ON h.casenodeid   = cd.casenodeid\r\n",
```
stef-solirius, line 280: Just an aesthetic note: should we not use camel case also for casenodeid > caseNodeId. Rest LGTM

Fred83200: Following the naming convention from the Horizon views which uses lower case, snake_case. If I change it to camelCase it would break the joins. Bear in mind everything will be renamed in HRM layer.

PR #2154, opened 2025-10-08
Feat/theodw 2354

Fred83200:
[https://pins-ds.atlassian.net/browse/THEODW-2354](url)

This PR introduces schema alignment and consistency improvements across the Harmonised (odw_harmonised_db.appeal_s78) and Curated (odw_curated_db.appeal_s78) layers for the Appeal S78 entity
It ensures that both layers are fully synchronised, schema-consistent, and reflect only the latest active data

Harmonised Layer Changes:
- Updated the Harmonised SCD2 build logic to align its schema with the Service Bus (sb_appeal_s78) schema
- The goal is to ensure that downstream layers (Curated) always inherit a consistent schema
- The following logic was added at the end of the Harmonised build process:
- Read the schema from the Service Bus source table (sb_schema)
- Added or cast columns in the Harmonised dataset (scd2) to match the Service Bus data types and order
- Explicitly projected columns in the same order as Service Bus
- Wrote the final Delta table partitioned by IsActive
- No changes were made to business logic or data selection this update affects only schema consistency and data type alignment

Curated Layer Changes
- Rebuilt odw_curated_db.appeal_s78 from the Harmonised FINAL
- Removed dependency on sb_appeal_s78 the Curated layer now sources data directly from the Harmonised layer
- Added logic to:
- Select only approved schema columns from Harmonised
- Tolerate harmless schema drift (ignore any missing optional fields)


workspace/notebook/py_sb_horizon_harmonised_appeal_s78.json
```diff
@@ -654,16 +654,30 @@
...
+					"    else:\r\n",
+					"        scd2 = scd2.withColumn(f.name, F.col(f.name).cast(f.dataType))\r\n",
+					"\r\n",
+					"# Here I project to the exact SB column order\r\n",
+					"final_cols = [f.name for f in sb_schema]\r\n",
+					"scd2_final = scd2.select(*[F.col(c) for c in final_cols])\r\n",
+					"\r\n",
 					"# And to finish I write final partitioned by activity\r\n",
-					"(scd2\r\n",
+					"(scd2_final\r\n",
```
stef-solirius, line 672: What's the name we are giving to the delta table?

workspace/notebook/py_sb_horizon_harmonised_appeal_s78.json
```diff
@@ -654,16 +654,30 @@
...
 					")\r\n",
 					"\r\n",
+					"# Here I Build the final column list from the Service Bus schema\r\n",
+					"sb_schema = spark.table(service_bus_table).schema\r\n",
+					"\r\n",
+					"# Here I Cast/add columns to match SB schema exactly\r\n",
+					"for f in sb_schema:\r\n",
+					"    if f.name not in scd2.columns:\r\n",
+					"        scd2 = scd2.withColumn(f.name, F.lit(None).cast(f.dataType))\r\n",
+					"    else:\r\n",
```
NHAL22, line 664: would it be necessary to record which columns are missing when there is a schema misalignment?

Fred83200: Good point - added the logic.

stef-solirius (approved):
LGTM, thanks Fred

PR #2444, opened 2026-04-07
Feat/theodw 2950 anonymise pins inspector

stef-solirius:
1. JIRA Ticket Reference  : https://pins-ds.atlassian.net/browse/THEODW-2950
 2. Anonymise pins-inspector table in dev and test Synapse workspaces for Casework integration testing
 3.  Summary
This PR implements automatic anonymisation of sensitive fields in the pins_inspector harmonised table for DEV and TEST environments only, enabling the Casework team to conduct integration testing without handling PII or sensitive data.

Key Changes

1. Anonymisation Integration (py_harmonised_pins_inspector.json)
•  Added imports for AnonymisationEngine, AnonymisationConfig, and Util
•  Integrated environment detection using Util.is_non_production_environment()
•  Fetches Purview classifications for the pins_inspector table
•  Applies anonymisation strategies before writing to harmonised layer
•  Uses sapId as seed column for deterministic anonymisation

2. AddressStrategy Enhancement (odw/core/anonymisation/base.py)
•  New strategy: AddressStrategy for address field anonymisation
•  Struct field support: Handles nested struct columns (e.g., address with addressLine1, addressLine2, etc.)
•  Selective redaction: Redacts address lines, town/city, and county but preserves postcode for geographic granularity
•  Classification support: Recognizes MICROSOFT.PERSONAL.PHYSICALADDRESS from Purview

4. Compatibility Fixes
•  Pydantic downgrade: Changed from v2.12.5 → v1.10.0 to avoid typing_extensions.Sentinel import error in Synapse
•  Lazy import: Moved DefaultAzureCredential import to avoid triggering dependency issues
•  Private function fix: Removed imports of private functions; credentials now fetched directly via mssparkutils

5. Policy Configuration (odw-config)
•  Added MICROSOFT.PERSONAL.PHYSICALADDRESS to the anonymisation policy allowlist
•  Ensures address fields are included in anonymisation scope

Testing
- Tested in DEV environment
- Confirmed anonymisation applies to 5 columns: firstName, lastName, email, inspectorManager
- Verified postcode preservation

image

image


odw/core/anonymisation/base.py
```diff
@@ -105,15 +105,14 @@ def apply(self, df: DataFrame, column: str, seed: Column, context: dict) -> Data
         """Mask the email local part using native Spark column expressions.
 
         Keeps the first and last character of the local part and replaces the
-        middle with '*'. The domain is preserved unchanged. Non-email strings
-        are masked end-to-end keeping only first and last character.
+        middle with '*'. The domain is always replaced with @pins.com.
+        Non-email strings are masked end-to-end keeping only first and last character.
         """
         col_expr = F.col(column).cast("string")
         has_at = col_expr.contains("@")
```
Fred83200: I'm thinking if this is not too weak as an email detector ? 
Cause if anything contains an @ it thinks it is an email

odw/core/anonymisation/base.py
```diff
@@ -105,15 +105,14 @@ def apply(self, df: DataFrame, column: str, seed: Column, context: dict) -> Data
...
         local_part = F.split(col_expr, "@").getItem(0)
-        domain_part = F.split(col_expr, "@").getItem(1)
         masked_local = F.regexp_replace(local_part, r"(?<=.).(?=.)", "*")
-        email_result = F.concat(masked_local, F.lit("@"), domain_part)
+        email_result = F.concat(masked_local, F.lit("@pins.com"))
         non_email_result = F.regexp_replace(col_expr, r"(?<=.).(?=.)", "*")
         result = F.when(F.col(column).isNull(), None).otherwise(F.when(has_at, email_result).otherwise(non_email_result))
```
Fred83200: This works I'd have maybe reused col_expr instead of calling F.col(column) just for readability

Fred83200: Something like : 
result = F.when(col_expr.isNull(), None).otherwise(
    F.when(has_at, email_result).otherwise(non_email_result)
)

odw/core/anonymisation/base.py
Fred83200: I'd maybe make your test cover some weird but possible values like : 
a@x.com
abc@
ab@x.com
@domain.com
not an email @ value

Things like this

odw/core/anonymisation/base.py
stef-solirius: Good observation, I've changed that.

odw/core/anonymisation/base.py
stef-solirius: Yeah very true, which makes me think I also need to change the unit_test to make it more resilient.

Fred83200 (dismissed):

Fred83200 (dismissed):

Fred83200 (dismissed):

Fred83200 (dismissed):

odw/core/util/util.py
```diff
@@ -44,9 +44,10 @@ def get_environment(cls) -> str:
     @classmethod
     def is_non_production_environment(cls) -> bool:
         """
-        Return True if the current environment is DEV only.
+        Return True if the current environment is DEV or TEST.
         """
-        return cls.get_environment() == "DEV"
+        env = cls.get_environment()
+        return env in ("DEV", "TEST")
```
HarrisonBoyleThomas: Should be extended for `BUILD` too. I'd also convert everything to `lower()` just in case

stef-solirius: You're right, just made the change

HarrisonBoyleThomas (approved):

PR #2446, opened 2026-04-09
Mpesc final changes am

Amerali786:
**PR Template**

 1. JIRA Ticket Reference  :
         **<[!-- Replace with JIRA ticket number and title --](https://pins-ds.atlassian.net/browse/HRP-482)>**
    [ HRP-482 Live migration tasks]

 2.  Summary of the work : 

    [  Implemented fixes and enhancements for Horizon → MPESC data mapping to align with audit findings.]

]
 
 3. Scripts have been tested end to end in dev from horizon into the dev mpesc sql server and data has displayed as expected on the front end of the mpesc system in dev.


workspace/notebook/Transform-Decision.json
```diff
@@ -505,23 +506,55 @@
...
+					"        \"decisionType\",\n",
+					"        F.when(\n",
+					"            (\n",
+					"                F.col(\"outcomeDate\").isNotNull()\n",
+					"            ) &\n",
+					"            (\n",
+					"                F.col(\"decisionType\").isNull() |\n",
+					"                F.lower(F.trim(F.col(\"decisionType\"))).isin(\"undefined\", \"null\", \"none\", \"\")\n",
+					"            ),\n",
+					"            F.lit(\"Other\")\n",
```
stef-solirius, line 530: This silently converts missing/invalid decision types to "Other" which could mask data quality issues. We're potentially losing visibility into bad source data

workspace/notebook/Transform-Decision.json
```diff
@@ -505,23 +506,55 @@
...
 					"        F.col(\"decisionBranch\"),\n",
 					"        F.col(\"casePublishFlag\"),\n",
 					"        F.col(\"dtypeId\")\n",
-					"    ).filter(\n",
-					"        F.col(\"outcomeRaw\").isNotNull() &\n",
-					"        (~F.lower(F.trim(F.col(\"outcomeRaw\"))).isin(\"undefined\", \"null\", \"none\", \"\"))\n",
+					"    )\n",
+					"    .filter(\n",
+					"        (\n",
+					"            F.col(\"outcomeRaw\").isNotNull() &\n",
```
stef-solirius, line 553: This OR condition means records with outcomeDate but no valid outcomeRaw will now pass through. Have you verified this doesn't introduce incomplete decision records?

workspace/notebook/Transform-Decision.json
```diff
@@ -505,23 +506,55 @@
 					"collapsed": false
 				},
 				"source": [
-					"decision_mapped = decision_mapped.withColumn(\"outcomeGuid\", F.expr(\"uuid()\"))\n",
+					"from pyspark.sql import functions as F\n",
 					"\n",
-					"decision_final = (\n",
+					"outcome_guid_map = (\n",
```
stef-solirius, line 511: Also, you're generating ONE outcomeGuid per caseId, meaning all decisions for the same case share the same outcome GUID. Is this intentional? Typically each decision/outcome should have a unique identifier...

workspace/notebook/Transform-Decision.json
```diff
@@ -319,12 +319,13 @@
 					"        F.lit(None).cast(\"string\").alias(\"attr1\"),\n",
 					"\n",
 					"        # CHANGE THESE COLUMN NUMBERS AFTER INSPECTION\n",
-					"        F.to_timestamp(clean_str(F.col(\"13\").cast(\"string\")), \"yyyy-MM-dd HH:mm:ss\").alias(\"decisionIssue\"),\n",
+					"        F.to_timestamp(clean_str(F.col(\"12\").cast(\"string\")), \"yyyy-MM-dd HH:mm:ss\").alias(\"decisionIssue\"),\n",
```
stef-solirius, line 322: Using numeric column references ("12", "13") is extremely fragile. If the source schema changes or columns are reordered, this breaks silently

workspace/pipeline/LOAD-DECISION.json
stef-solirius, line 1: "Load Decision" activity now depends on "Load Outcome" completing first. Both activities read from the SAME source folder (MPESC-TRANSFORM/Decision). Why does Outcome need to load first? Is there a FK constraint?

workspace/notebook/Transform-Decision.json
```diff
@@ -42,11 +42,11 @@
 					"type": "AAD",
 					"authResource": "https://dev.azuresynapse.net"
 				},
-				"sparkVersion": "3.5",
+				"sparkVersion": "3.4",
```
stef-solirius, line 45: We shouldn't downgrade the sparkVersion > please revert to 3.5

Amerali786: I believe it is using the correct one from going into it on synapse but have recommitted with 3.5

workspace/pipeline/LOAD-DECISION.json
Amerali786: Yes you need a outcome to have a decision.  Decision uses the outcome guid as a reference. An outcome can have many decisions. We have outcome guid created as part of transform which upserts into the outcome table so it doesnt duplicate any outcomes

workspace/notebook/Transform-Decision.json
Amerali786: Ye i understand what you are saying. The column references are reffering to attribute IDs which i have got by pivoting a attribute data table from horizon. Ive created a data dictionary here (not sure if you can access): https://pinso365.sharepoint.com/:x:/r/sites/HorizonMigrationReview/Shared%20Documents/Alpha/Data%20Migration/Horizon%20Data/Horizon%20Data%20Tables.xlsx?d=w30cf01d390694f158180622ef60a6102&csf=1&web=1&e=hTGutd

In this step i do give the attribute a alias to identify going forward in the notebook.

workspace/notebook/Transform-Decision.json
Amerali786: My intention in my new decision notebook is to generate 1 outcome guid per case and that links to many decisions. Generally in this data set there is max 2 decisions which i draw from 3 different sources. I believe it shows up as expected when it was tested.

workspace/notebook/Transform-Decision.json
Amerali786: I believe this was intentional. This was a decision made by the BAs from discussion with the client that where they have a closed date but no decision we add in Other as the decision from what I remember.

workspace/notebook/Transform-Decision.json
Amerali786: Ye this is the step where we apply other to decisions with a outcome date but no outcome value. This was a decision by my BA to include as part of the migration

stef-solirius (approved):
In my review I flagged that the import process lacks sufficient validation, specifically recommending null counts, row counts and schema validation to catch any data loss or corruption during the copy activity. The author acknowledged this is fair but noted it can't be added within the scope of this PR. They do have row counts in place for certain key tables (case officer and inspector notebooks), but schema validation and null counts are not currently implemented.

We accepted this as a known risk and recommend a plan is in place for future improvement. We'll now go ahead and approve.

PR #2448, opened 2026-04-10
Feat/theodw 2985 fix curated mi pins notebooks

anushaveerapalli68:
JIRA Ticket Reference: https://pins-ds.atlassian.net/browse/THEODW-2985

- Updated curated‑Mipins notebooks to preserve external table behaviour during overwrite operations.
- Retrieved the existing table path using DESCRIBE EXTENDED <table_name>.
- Extracted the Location value and stored it as table_path.
- Added .option("path", table_path) to the DataFrame write operation.
- Ensures Spark writes back to the existing external directory instead of creating a managed table.
- Prevents curated tables from being unintentionally recreated as managed tables.
- Curated tables now consistently remain external with their original storage paths retained.

Files Created/Modified:

Pipeline: workspace/pipeline/Releases/33.0.0/rel_33_0_0_convert_migration_tables.json
◦ Edited the Post-deployment pipeline that runs the py_cur_mig_convert_managed_to_external notebook that includes all curated tables and curated MiPINS tables

◦ Configured with all 6 MiPINS tables:

-    odw_curated_db.appeal_service_user_curated_mipins
-     odw_curated_db.appeal_event_curated_mipins
-     odw_curated_db.appeals_has_curated_mipins
-     odw_curated_db.entraid_curated_mipins
-     odw_curated_db.appeals_s78_curated_mipins
-     odw_curated_db.appeal_event_estimate_curated_mipins


workspace/notebook/appeal_event_curated_mipins.json
```diff
@@ -110,65 +110,64 @@
...
-						"transient": {
-							"deleting": false
-						}
-					},
-					"collapsed": false
-				},
-				"source": [
+					"from pyspark.sql import SparkSession\n",
+					" \n",
+					"spark = SparkSession.builder.getOrCreate()\n",
```
HarrisonBoyleThomas, line 121: I would move these to the top of the notebook

HarrisonBoyleThomas: Ignore me, just noticed the whole notebook is SQL

workspace/notebook/appeal_event_curated_mipins.json
```diff
@@ -110,65 +110,64 @@
...
+					" \n",
+					"# Get existing external table location\n",
+					"df_table_path = spark.sql(f\"DESCRIBE EXTENDED {table_name}\")\n",
+					" \n",
+					"table_path = (\n",
+					"    df_table_path\n",
+					"        .filter(\"col_name = 'Location'\")\n",
+					"        .select(\"data_type\")\n",
+					"        .collect()[0][0]\n",
+					")\n",
```
HarrisonBoyleThomas, line 133: How would this handle situations where the table does not exist beforehand?

workspace/notebook/appeal_event_estimate_curated_mipins.json
```diff
@@ -121,14 +121,30 @@
...
+					"\n",
+					"# Get existing external table location\n",
+					"df_table_path = spark.sql(f\"DESCRIBE EXTENDED {table_name}\")\n",
+					"\n",
+					"table_path = (\n",
+					"    df_table_path\n",
+					"        .filter(\"col_name = 'Location'\")\n",
+					"        .select(\"data_type\")\n",
+					"        .collect()[0][0]\n",
+					")\n",
```
HarrisonBoyleThomas, line 138: Same here

workspace/notebook/appeal_event_curated_mipins.json
anushaveerapalli68: This approach assumes the table already exists. If the table doesn't exist, Describe extended will fail because there is no meta data to retrieve the location from. In this case, the expectation is that the curated table is pre created as an external table.

PR #2452, opened 2026-04-13
Feat/theodw 2968 appeal representation

HarrisonBoyleThomas:
Ticket: https://pins-ds.atlassian.net/browse/THEODW-2968

Add unit and integration tests for appeal representation curation process


odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
```diff
@@ -0,0 +1,301 @@
...
+                    False,
+                    None,
+                    None,
+                    [],
+                    [],
+                    "lpa",
+                    None,
+                    "statement",
+                    datetime(2025, 1, 1, 0, 0, 0, 0).isoformat(),
+                    ["aaaaa"],
```
Fred83200, line 199: The legacy notebook selects documentIds directly from odw_harmonised_db.sb_appeal_representation, but these expected values look like they are coming from message_id instead.
I'm thinking for parity with the legacy notebook the curated documentIds should match the harmonised documentIds field not message_id

odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
```diff
@@ -0,0 +1,301 @@
...
+                    False,
+                    None,
+                    None,
+                    [],
+                    [],
+                    "lpa",
+                    None,
+                    "statement",
+                    datetime(2025, 1, 2, 0, 0, 0, 0).isoformat(),
+                    ["bbbbb"],
```
Fred83200, line 216: Same for this one

odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
```diff
@@ -0,0 +1,301 @@
...
+                    False,
+                    None,
+                    None,
+                    [],
+                    [],
+                    "lpa",
+                    None,
+                    "statement",
+                    datetime(2025, 1, 4, 0, 0, 0, 0).isoformat(),
+                    ["ddddd"],
```
Fred83200, line 233: and same for this one

odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
```diff
@@ -0,0 +1,301 @@
...
+                ),
+            ),
+            schema=self.generate_curated_data_schema(),
+        )
+
+    def test__appeal_representation_curation_process__run__with_existing_data(self):
+        """
+        - Given I a have an existing appeal representation table and some new appeal representation data in the harmonised layer
+        - When I run the appeal representation curation process
+        - Then the new data should be added to the existing table
```
Fred83200: Just wording typo the legacy notebook writes with mode(overwrite) so this is overwrite behaviour rathen than add to existing table

odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
```diff
@@ -0,0 +1,301 @@
...
+        self.write_existing_table(
+            spark,
+            existing_harmonised_data,
+            "sb_test_arcp_r_ed",
+            "odw_harmonised_db",
+            "odw-harmonised",
+            "sb_appeal_representation",
+            "overwrite",
+        )
+        existing_curated_data = spark.createDataFrame([], self.generate_curated_data_schema())
```
Fred83200: This creates an empty curated table so the test will pass wether it overwrites or append. As the legacy notebook uses overwrite maybe you could seed this table with one or two stale rows first and then assert the active HRM rows remain after the run()

odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
```diff
@@ -0,0 +1,301 @@
...
+                    "ODT",
+                    5,
+                    datetime(2025, 1, 3, 0, 0, 0, 0).isoformat(),
+                    None,
+                    None,
+                    "N",  # This entry should be dropped by the curation process
+                    "ccccc",
+                )
+            ],
+            schema=T.StructType(
```
Fred83200, line 129: Some values look inconsistent with the declared schema, Might need to sanity check it

odw/test/unit_test/etl/curated/test_appeal_representation_curation_process.py
```diff
@@ -0,0 +1,196 @@
...
+                    [
+                        T.StructField("col_name", T.StringType(), False),
+                        T.StructField("data_type", T.StringType(), False),
+                        T.StructField("comment", T.StringType(), True),
+                    ]
+                ),
+            ),
+        ]
+        expected_output = {"harmonised_data": spark_sql_side_effects[0], "curated_data_description": spark_sql_side_effects[1]}
+        with mock.patch.object(SparkSession, "sql", side_effect=spark_sql_side_effects):
```
Fred83200: This check the returned DF which is good but maybe you can assert the SQL here ? Cause the notebook depends on the SELECT .... FROM sb_appeal_representation WHERE IsActive = 'Y' and also the DESCRIBE EXTENDED on the curated table. I think asserting those queries would be good for parity

odw/test/unit_test/etl/curated/test_appeal_representation_curation_process.py
```diff
@@ -0,0 +1,196 @@
...
+                actual_keys = set(actual_output.keys())
+                assert expected_keys == actual_keys, (
+                    f"Expected a dictionary with keys {expected_keys} to be returned by load_data(), but received the keys {actual_keys} instead"
+                )
+                assert_dataframes_equal(expected_output["harmonised_data"], actual_output["harmonised_data"])
+                assert_dataframes_equal(expected_output["curated_data_description"], actual_output["curated_data_description"])
+
+    def test__appeal_representation__process(self):
+        spark = PytestSparkSessionUtil().get_spark_session()
+        mock_data = {
```
Fred83200, line 168: It looks here that the input DF is already very close to the curated shape, might be worth adding a row with IsActive = 'N' and some extra HRM only columns so your test proves the same filter behavior as the notebook not just the write metadata

odw/test/unit_test/etl/curated/test_appeal_representation_curation_process.py
```diff
@@ -0,0 +1,196 @@
...
+                    "container_name": "odw-curated",
+                    "blob_path": "appeal_representation",
+                    "file_format": "parquet",
+                    "write_mode": "overwrite",
+                    "write_options": {},
+                }
+            }
+            with mock.patch.object(AppealRepresentationCurationProcess, "__init__", return_value=None):
+                inst = AppealRepresentationCurationProcess()
+                actual_data_to_write, _ = inst.process(mock_data)
```
Fred83200, line 254: The notebook does a straight SELECT of 15 columns so checking that exact curated projection would make your test more robust I think

odw/test/unit_test/etl/curated/test_appeal_representation_curation_process.py
HarrisonBoyleThomas: I wanted to avoid asserting against the query because there might be slight variations in structure

odw/test/unit_test/etl/curated/test_appeal_representation_curation_process.py
HarrisonBoyleThomas: I'm not quite sure about this one. I want to avoid dropping the `where` clause in `load_data`, since this might impact performance. I also wanted to avoid doing `select *` for the same reason. The query itself is quite small since it just copies the active rows from harmonised->curated, so there shouldn't be any xform logic in `process`

odw/test/unit_test/etl/curated/test_appeal_representation_curation_process.py
HarrisonBoyleThomas: Same reasoning as prev. point

odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
HarrisonBoyleThomas: Updated

odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
HarrisonBoyleThomas: Good spot, updated

odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
HarrisonBoyleThomas: Good spot, updated

odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
HarrisonBoyleThomas: Updated

odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
HarrisonBoyleThomas: Updated

odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
HarrisonBoyleThomas: Updated

odw/test/unit_test/etl/curated/test_appeal_representation_curation_process.py
Fred83200: Maybe we can avoid asserting against the full query but the test does not prove that load_data() is reading the correct source and doing the correct filter. 
Maybe we can make sure it does ? 
I was thinking maybe doing an assert on important bits only for example that one query targets sb_appeal_representation, one targets DESCRIBE EXTENDED and that the active row filter is still part of the load

odw/test/unit_test/etl/curated/test_appeal_representation_curation_process.py
Fred83200: That's fair - Happy with it

odw/test/unit_test/etl/curated/test_appeal_representation_curation_process.py
Fred83200: Fair too - let's leave it as is

odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
Fred83200: That's much better - Just one thing you should double check that the seeded existing curated rows match the curated schema exactly as it looks more like the harmonised shaped rows atm

HarrisonBoyleThomas: I just double checked the curated table, looks like it matches to me?

odw/test/unit_test/etl/curated/test_appeal_representation_curation_process.py
HarrisonBoyleThomas: I've updated the test to create database tables instead of mocking the return value of `spark.sql`

odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
```diff
@@ -0,0 +1,339 @@
...
+                T.StructField("documentIds", T.ArrayType(T.StringType(), True), True),
+            ]
+        )
+
+    def generate_curated_data(self):
+        spark = PytestSparkSessionUtil().get_spark_session()
+        return spark.createDataFrame(
+            data=(
+                (
+                    2,
```
Fred83200, line 185: I think representationID > StringType
caseId > LongType
caseReference > StringType

Spakr might maybe coerce those not sure but for clarity I'll maybe change it 
here to "2"

odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
```diff
@@ -0,0 +1,339 @@
...
+                    3,
+                    603,
```
Fred83200, line 187: This one to "603"

odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
```diff
@@ -0,0 +1,339 @@
...
+            "odw_harmonised_db",
+            "odw-harmonised",
+            "sb_appeal_representation",
+            "overwrite",
+        )
+        # Existing data should be overwritten
+        existing_curated_data = spark.createDataFrame(
+            [
+                (
+                    11,
```
Fred83200, line 260: Same here "11"

odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
```diff
@@ -0,0 +1,339 @@
...
+                    12,
+                    612,
```
Fred83200, line 262: And this one "612"

odw/test/unit_test/etl/curated/test_appeal_representation_curation_process.py
HarrisonBoyleThomas: I've updated the mock data to have a row where `IsActive=N`

odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
HarrisonBoyleThomas: Updated

odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
HarrisonBoyleThomas: Updated

odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
HarrisonBoyleThomas: Updated

odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
HarrisonBoyleThomas: Updated

odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
```diff
@@ -0,0 +1,339 @@
...
+                    "lpa",
+                    None,
+                    "statement",
+                    datetime(2025, 1, 2, 0, 0, 0, 0).isoformat(),
+                    ["bbbbb"],
+                ),
+                (
+                    6,
+                    7,
+                    607,
```
Fred83200: I think this now needs to be a string too like you did above

odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
```diff
@@ -0,0 +1,339 @@
...
+            1,
+            "ODT",
+            5,
+            datetime(2025, 1, 4, 0, 0, 0, 0).isoformat(),
+            "Y",
+            "ddddd",
+        ),
+    ]
+
+    def generate_harmonised_table(self):
```
Fred83200, line 97: I think here it's the same - 
T.StructField("migrated", T.StringType(), True),
T.StructField("SourceSystemID", T.StringType(), True)

But the data you pass are integers like : 1, 5

odw/test/unit_test/etl/curated/test_appeal_representation_curation_process.py
```diff
@@ -0,0 +1,283 @@
...
+                expected_data_to_write_without_data = {k: {sk: sv for sk, sv in v.items() if sk != "data"} for k, v in expected_data_to_write.items()}
+                actual_data_to_write_without_data = {k: {sk: sv for sk, sv in v.items() if sk != "data"} for k, v in actual_data_to_write.items()}
+                assert expected_data_to_write_without_data == actual_data_to_write_without_data
+                assert (
+                    expected_data_to_write["odw_curated_db.appeal_representation"]["data"]
+                    == actual_data_to_write["odw_curated_db.appeal_representation"]["data"]
```
Fred83200: Not sure about this "==". For Spark DF this is not equality assertion in the same was as python obj. 
You should just have it like this : 
assert_dataframes_equal(
    expected_data_to_write["odw_curated_db.appeal_representation"]["data"],
    actual_data_to_write["odw_curated_db.appeal_representation"]["data"],
)

odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
```diff
@@ -0,0 +1,339 @@
...
+                    datetime(2025, 1, 3, 0, 0, 0, 0).isoformat(),
+                    None,
+                    None,
+                    "N",  # This entry should be dropped by the curation process
+                    "ccccc",
+                )
+            ],
+            schema=T.StructType(
+                [
+                    T.StructField("appealRepresentaionID", T.LongType(), True),
```
Fred83200: This one looks mispelled , shouldn't it be : appealRepresentationID

HarrisonBoyleThomas: Lemme double check, i copied this from synapse

odw/test/unit_test/etl/curated/test_appeal_representation_curation_process.py
HarrisonBoyleThomas: Good spot, updated

odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
HarrisonBoyleThomas: Updated

odw/test/integration_test/etl/curated/test_appeal_representation_curation_process.py
HarrisonBoyleThomas: Updated

Fred83200 (approved):

PR #2454, opened 2026-04-14
Adding unit/int tests for listed_building

Fred83200:
[THEODW-2969](https://pins-ds.atlassian.net/browse/THEODW-2969)

This PR introduces the full test coverage and transformation scaffolding for the listed_building entity across all layers (Standardised > Harmonised > Curated)

- Created template transformation files for each layer to align with the ETL structure
- Implemented unit and integration tests
- All tests are currently marked as xfail until the actual transformation logic is implemented

Coverage includes:

- Column mapping and schema validation
- CDC behaviour (row versioning, updates, inserts) in Harmonised
- Safe merge logic in Curated (insert/update detection, schema alignment, edge cases)
- Handling of nulls, duplicates, and empty datasets
- Consistency with legacy notebook behaviour

[THEODW-2969]: https://pins-ds.atlassian.net/browse/THEODW-2969?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


odw/test/integration_test/etl/curated/test_listed_building_curated_process.py
```diff
@@ -0,0 +1,403 @@
...
+
+        data_to_write = mock_write.call_args[0][0]
+        df = data_to_write[inst.OUTPUT_TABLE]["data"]
+
+        assert df.count() == 1
+        assert df.where(F.col("entity") == 1001).count() == 1
+        assert result.metadata.insert_count == 0
+        assert result.metadata.update_count == 0
+
+    def test__listed_building_curated_process__run__drops_duplicate_source_rows_and_filters_inactive_rows(
```
stef-solirius: The test name says 'filters inactive rows' but _duplicate_harmonised_df only contains rows where isActive='Y' — no inactive rows are present in this fixture. Inactive filtering is already covered separately in test__initial_load_matches_legacy. Either extend the fixture to include an inactive row so both behaviours are tested together, or simplify the name to test__drops_duplicate_source_rows to match what it actually asserts.

odw/test/integration_test/etl/harmonised/test_listed_building_harmonisation_process.py
```diff
@@ -0,0 +1,538 @@
...
+                "grade-ii",
+                "https://example.com/lb-001",
+                "II",
+            ),
+        ],
+        schema=_standardised_schema(),
+    )
+
+
+def _identical_standardised_df(spark):
```
stef-solirius: `_identical_standardised_df` is a confusing name here — the data it contains has `name="Old Building One"`, which happens to match what is already in `_existing_harmonised_df`. The "identical" refers to the post-harmonisation state, not to the standardised data. A clearer name like `_unchanged_standardised_df` or `_standardised_df_matching_existing_harmonised` would make the test intent obvious to a reader without needing to trace through the fixture.

odw/test/integration_test/etl/harmonised/test_listed_building_harmonisation_process.py
```diff
@@ -0,0 +1,538 @@
...
+            with (
+                mock.patch.object(inst, "load_data", return_value=source_data),
+                mock.patch.object(inst, "write_data") as mock_write,
+            ):
+                result = inst.run()
+
+        data_to_write = mock_write.call_args[0][0]
+        df = data_to_write[inst.OUTPUT_TABLE]["data"]
+
+        assert df.where((F.col("entity") == "1001") & (F.col("isActive") == "Y")).count() == 1
```
stef-solirius, line 537: The assertions here say entity `1001` stays `isActive="Y"` even though entity `9999` has taken over reference `LB-001`. Is this intentional legacy behaviour — i.e. the harmonisation join is keyed on `entity` not `reference`, so entity `1001` is never touched? If so, this is a non-obvious edge case worth a short comment in the test to explain the expected behaviour explicitly so the implementer does not mistakenly treat it as a bug.

odw/test/unit_test/etl/curated/test_listed_building_curated_process.py
```diff
@@ -0,0 +1,396 @@
...
+        assert df.where(F.col("entity") == 1001).collect()[0]["name"] == "Building One Updated"
+        assert result.metadata.insert_count == 0
+        assert result.metadata.update_count == 1
+
+    def test__listed_building_curated_process__process__ignores_source_only_columns_during_schema_alignment(
+        self,
+    ):
+        spark = PytestSparkSessionUtil().get_spark_session()
+
+        source_df = spark.createDataFrame(
```
stef-solirius, line 422: The source DataFrame here is built with hardcoded positional column names rather than `_harmonised_schema()`. If the harmonised schema changes (e.g. a new column is added), this test could silently diverge from the real schema shape. Consider using `_harmonised_schema()` with an appended extra column instead, to keep the fixture consistent with the rest of the test file.

odw/test/unit_test/etl/harmonised/test_listed_building_harmonisation_process.py
```diff
@@ -0,0 +1,522 @@
...
+                existing_rows,
+                schema=_harmonised_schema(),
+            ),
+            "target_exists": True,
+        }
+
+        with mock.patch("odw.core.etl.transformation.harmonised.listed_building_harmonisation_process.LoggingUtil"):
+            inst = ListedBuildingHarmonisationProcess(spark)
+            data_to_write, result = inst.process(source_data=source_data)
+
```
stef-solirius, line 487: Same question as raised on the integration test counterpart — the assertions confirm entity `1001` remains `isActive="Y"` after entity `9999` takes over reference `LB-001`. If this is deliberate legacy behaviour (join is on `entity`, not `reference`), a short inline comment explaining why entity 1001 is untouched would save the implementer from having to second-guess it.

odw/test/integration_test/etl/curated/test_listed_building_curated_process.py
Fred83200: Agree- name changed

odw/test/integration_test/etl/harmonised/test_listed_building_harmonisation_process.py
Fred83200: Agree changed the naming to _unchanged_standardised_df

odw/test/integration_test/etl/harmonised/test_listed_building_harmonisation_process.py
Fred83200: Agree , Added comments before the asserts

odw/test/unit_test/etl/curated/test_listed_building_curated_process.py
Fred83200: Agree I added a _harmonised_schema_with_extra_source_column helper which i build the src from. 
also added the get_spark_session.
I also added the target_df to use _curated_schema() for consistency

odw/test/unit_test/etl/harmonised/test_listed_building_harmonisation_process.py
Fred83200: Added the same comment there too

stef-solirius (approved):

PR #2456, opened 2026-04-15
Adding nsip_invoice unit/int tests

Fred83200:
[THEODW-2990](https://pins-ds.atlassian.net/browse/THEODW-2990)

Implemented full test coverage for the nsip_invoice entity across harmonised and curated layers for the ETL refactor

Added unit tests to validate transformation logic in isolation:
- Array explosion (invoices)
- Filtering rules (null and empty handling)
- Surrogate key generation
- IsActive / ValidTo logic
- Migrated flag behaviour
- RowID hashing aligned with legacy logic

Added integration tests to validate end to end ETL behaviour:
- Initial load vs incremental load scenarios
- Ingestion date filtering
- Active/inactive record transitions
- Duplicate handling
- Preservation of unrelated existing data
- Curated layer filtering (IsActive = 'Y') and output shape

All tests are currently marked with xfail as the transformation logic has not yet been implemented.

[THEODW-2990]: https://pins-ds.atlassian.net/browse/THEODW-2990?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


odw/test/integration_test/etl/curated/test_nsip_invoice_curated_process.py
```diff
@@ -0,0 +1,310 @@
...
+                caseId=2003,
+                caseReference="EN010003",
+                invoiceNumber="INV-003",
+                amountDue="300.00",
+                IsActive="N",
+            ),
+        ]
+
+        source_data = {
+            "source_data": spark.createDataFrame(source_rows, schema=_harmonised_schema()),
```
HarrisonBoyleThomas: I don't think this will work, the return value of `load_data` will overwrite the `source_data` property when `process` is called. You will need to write the dataframe to a harmonised table/parquet (whatever the notebook does)

HarrisonBoyleThomas: This is the case for all of the integration tests

odw/test/integration_test/etl/curated/test_nsip_invoice_curated_process.py
```diff
@@ -0,0 +1,310 @@
...
+            "target_exists": False,
+        }
+
+        with (
+            mock.patch("odw.core.etl.etl_process.LoggingUtil") as mock_etl_logging,
+            mock.patch("odw.core.etl.transformation.curated.nsip_invoice_curated_process.LoggingUtil") as mock_process_logging,
```
HarrisonBoyleThomas, line 190: You might also need to patch `Util.get_storage_account` since it will probably be called by `process`

odw/test/integration_test/etl/curated/test_nsip_invoice_curated_process.py
```diff
@@ -0,0 +1,310 @@
...
+        ):
+            mock_etl_logging.return_value = mock.Mock()
+            mock_process_logging.return_value = mock.Mock()
+
+            inst = NsipInvoiceCuratedProcess(spark)
+
+            with (
+                mock.patch.object(inst, "load_data", return_value=source_data),
+                mock.patch.object(inst, "write_data") as mock_write,
+            ):
```
HarrisonBoyleThomas: Similar to my previous comment, i don't think we should mock these since it could mask issues. This would be better as a unit test imo

odw/test/integration_test/etl/curated/test_nsip_invoice_curated_process.py
```diff
@@ -0,0 +1,310 @@
...
+        data_to_write = mock_write.call_args[0][0]
+        df = data_to_write[inst.OUTPUT_TABLE]["data"]
+
+        assert data_to_write[inst.OUTPUT_TABLE]["write_mode"] == "overwrite"
+        assert df.count() == 2
+        assert result.metadata.insert_count == 2
+        assert result.metadata.update_count == 0
+        assert df.where(F.col("invoiceNumber") == "INV-001").count() == 1
+        assert df.where(F.col("invoiceNumber") == "INV-002").count() == 1
+        assert df.where(F.col("invoiceNumber") == "INV-003").count() == 0
```
HarrisonBoyleThomas: I think it would be good to squash these assertions because it can be tricky to identify the exact cause of a failure. If you create an "expected dataframe" and just assert that they match it should cover all the assertions in one line. There's a few examples in the other int tests

HarrisonBoyleThomas: e.g if there was an extra row in the data, then `df.count()` will fail, but to identify the cause we'll have to go into the code and debug it. If the assertion was against the full data we would be able to inspect the output without needing to run any extra code/debugging

odw/test/unit_test/etl/curated/test_nsip_invoice_curated_process.py
```diff
@@ -0,0 +1,326 @@
...
+        "amountDue": "100.50",
+        "paymentDueDate": "2025-01-20",
+        "invoicedDate": "2025-01-10",
+        "paymentDate": None,
+        "refundCreditNoteNumber": None,
+        "refundAmount": None,
+        "refundIssueDate": None,
+        "ODTSourceSystem": "ODT",
+        "SourceSystemID": "SRC-1",
+        "IngestionDate": "2025-02-01T10:00:00.000000+0000",
```
HarrisonBoyleThomas, line 82: I think this probably needs to be a datetime? Based on another entity i saw this col was a datetime

Fred83200: I checked the harmonised table schema for sb_nsip_invoice and IngestionDate is stored as varchar, so the test value is aligned with the actual data model. I’ll keep it as string to match the current behaviour

odw/test/integration_test/etl/curated/test_nsip_invoice_curated_process.py
Fred83200: Done

odw/test/integration_test/etl/curated/test_nsip_invoice_curated_process.py
Fred83200: Done

odw/test/integration_test/etl/curated/test_nsip_invoice_curated_process.py
Fred83200: Done

odw/test/integration_test/etl/curated/test_nsip_invoice_curated_process.py
Fred83200: Done

odw/test/integration_test/etl/curated/test_nsip_invoice_curated_process.py
```diff
@@ -0,0 +1,381 @@
...
+            StructField("refundIssueDate", StringType(), True),
+            StructField("ODTSourceSystem", StringType(), True),
+            StructField("SourceSystemID", StringType(), True),
+            StructField("IngestionDate", StringType(), True),
+            StructField("ValidTo", StringType(), True),
+            StructField("RowID", StringType(), True),
+            StructField("IsActive", StringType(), True),
+            StructField("Migrated", IntegerType(), True),
+        ]
+    )
```
HarrisonBoyleThomas, line 46: I think this should be more like this? (Raw snapshot from the notebook)

```
StructType([StructField('NSIPProjectInfoInternalID', LongType(), True), StructField('caseId', LongType(), True), StructField('caseReference', StringType(), True), StructField('invoices', ArrayType(StructType([StructField('invoiceStage', StringType(), True), StructField('invoiceNumber', StringType(), True), StructField('amountDue', DoubleType(), True), StructField('paymentDueDate', StringType(), True), StructField('invoicedDate', StringType(), True), StructField('paymentDate', StringType(), True), StructField('refundCreditNoteNumber', StringType(), True), StructField('refundAmount', DoubleType(), True), StructField('refundIssueDate', StringType(), True)]), True), True), StructField('ODTSourceSystem', StringType(), True), StructField('SourceSystemID', StringType(), True), StructField('IngestionDate', StringType(), True)])

```

odw/test/integration_test/etl/curated/test_nsip_invoice_curated_process.py
```diff
@@ -0,0 +1,381 @@
...
+            StructField("caseReference", StringType(), True),
+            StructField("invoiceStage", StringType(), True),
+            StructField("invoiceNumber", StringType(), True),
+            StructField("amountDue", FloatType(), True),
+            StructField("paymentDueDate", StringType(), True),
+            StructField("invoicedDate", StringType(), True),
+            StructField("paymentDate", StringType(), True),
+            StructField("refundCreditNoteNumber", StringType(), True),
+            StructField("refundAmount", FloatType(), True),
+            StructField("refundIssueDate", StringType(), True),
```
HarrisonBoyleThomas: Similar here, `sb_nsip_invoice` has this schema (if i inspect the schema in synapse)
```
StructType([StructField('NSIPInvoiceID', LongType(), True), StructField('caseId', LongType(), True), StructField('caseReference', StringType(), True), StructField('invoiceStage', StringType(), True), StructField('invoiceNumber', StringType(), True), StructField('amountDue', DoubleType(), True), StructField('paymentDueDate', StringType(), True), StructField('invoicedDate', StringType(), True), StructField('paymentDate', StringType(), True), StructField('refundCreditNoteNumber', StringType(), True), StructField('refundAmount', DoubleType(), True), StructField('refundIssueDate', StringType(), True), StructField('Migrated', IntegerType(), True), StructField('ODTSourceSystem', StringType(), True), StructField('SourceSystemID', StringType(), True), StructField('IngestionDate', StringType(), True), StructField('ValidTo', StringType(), True), StructField('RowID', StringType(), True), StructField('IsActive', StringType(), True), StructField('NSIPProjectInfoInternalID', LongType(), True)])

```

Fred83200: Should be good now

odw/test/integration_test/etl/curated/test_nsip_invoice_curated_process.py
Fred83200: Should be good now

HarrisonBoyleThomas (approved):

PR #2461, opened 2026-04-16
fix: resolve specialism accumulation bug in pins_inspector pipeline

stef-solirius:
## Summary

- Fixed syntax errors in `py_get_delta_table_changes` (`quarantine_folder` typo, `json.dumps` typos, `current_batch` typo, `oversized_count` variable inconsistency, invalid f-string expression, wrong argument passed to `get_json_size_bytes`)
- Fixed root cause of oversized inspector records in `py_harmonised_pins_inspector`: added `DISTINCT` inside `insp_spec_dedup` CTE to prevent specialism row accumulation across pipeline runs inflating a single inspector's `specialisms` array (observed: 14,828 entries for one inspector)
- Added `addr_dedup` CTE to deduplicate address rows per inspector before the join, preventing row fan-out if an inspector has multiple historical address records

## Test plan

- [ ] Run `py_harmonised_pins_inspector` and verify no inspector record has an unexpectedly large `specialisms` array
- [ ] Run `py_get_delta_table_changes` against `odw_curated_db.pins_inspector` and verify batch files are written without runtime errors
- [ ] Confirm oversized records are correctly routed to the quarantine folder rather than crashing the notebook
- [ ] Verify batch files are within the 200KB Service Bus target size

https://pins-ds.atlassian.net/browse/THEODW-2999


workspace/notebook/py_get_delta_table_changes.json
```diff
@@ -472,9 +472,10 @@
 				},
 				"source": [
 					"# Batching configuration\n",
-					"TARGET_BATCH_SIZE_BYTES = 256 * 1024  # 256KB target for Service Bus message size limit\n",
+					"TARGET_BATCH_SIZE_BYTES = 200 * 1024  # 256KB target for Service Bus message size limit\n",
```
HarrisonBoyleThomas: The comment needs to be updated. To `200KiB`

workspace/notebook/py_harmonised_pins_inspector.json
```diff
@@ -254,6 +259,16 @@
 					"                FROM {entraid_table}\n",
 					"                WHERE isActive = 'Y'\n",
 					"            ),\n",
+					"            -- Deduplicate address records: one row per inspector\n",
+					"            addr_dedup AS (\n",
+					"                SELECT *\n",
+					"                FROM (\n",
+					"                    SELECT *,\n",
+					"                        ROW_NUMBER() OVER (PARTITION BY StaffNumber ORDER BY StaffNumber) AS rn\n",
```
Fred83200: Is there anything better for the ordering ? like a latest timestamp , valid from or validto or something ? 
Just thinking that here it dedupes to one row but is it really deterministic ? unless Spark stabilise it maybe ?

workspace/notebook/py_harmonised_pins_inspector.json
```diff
@@ -239,8 +241,11 @@
 					"                            'validFrom', ValidFrom\n",
 					"                        )\n",
 					"                    ) AS specialisms\n",
-					"                FROM {inspector_specialisms_table}\n",
-					"                WHERE Current = 1\n",
+					"                FROM (\n",
+					"                    SELECT DISTINCT StaffNumber, QualificationName, Proficien, ValidFrom\n",
```
Fred83200: Are those the fields to define uniqueness ? for a specialism ?

workspace/notebook/py_get_delta_table_changes.json
```diff
@@ -532,14 +569,36 @@
...
+					"    else:\n",
+					"        file_path = f\"{target_folder}/batch_{idx:05d}.json\"\n",
+					"        mssparkutils.fs.put(file_path, batch_json, True)\n",
 					"\n",
 					"batch_count = len(batches)\n",
-					"try:\n",
-					"    print(f\"Successfully wrote {changes_count} changes across {batch_count} batch files\")\n",
-					"except Exception as e:\n",
-					"    print(f\"Failed to log info: {e}\")"
+					"print(f\"Successfully wrote {changes_count} changes across {batch_count} batch files\")\n",
```
Fred83200, line 597: I think here you can add just before that print : 
changes_count = len(rows_list)

Just realised that the changes_count is not aligned with what we actually processed in the batching step. So you'd get exactly how many changes on how many batch file

workspace/notebook/py_get_delta_table_changes.json
stef-solirius: good point, thanks

workspace/notebook/py_get_delta_table_changes.json
stef-solirius: Thanks, addressed

workspace/notebook/py_harmonised_pins_inspector.json
stef-solirius: On reflection , I've added also Proficien and ValidFrom

workspace/notebook/py_harmonised_pins_inspector.json
stef-solirius: You're right, changed fro IngestionDate

Fred83200 (approved):

PR #2468, opened 2026-04-20
Adding unit/int tests appeal_document

Fred83200:
[THEODW-2963](https://pins-ds.atlassian.net/browse/THEODW-2963)

Implemented unit and integration tests for appeal_document harmonised and curated refactor work.

This PR adds the new test files and process templates, based on the legacy Synapse notebook behaviour, so the upcoming refactor can be validated against the current output and logic. The tests are intentionally aligned to legacy behaviour, including existing quirks, to make sure the refactor matches what the notebooks do today.

The new tests are currently marked xfail until the harmonised and curated process implementations are completed.

[THEODW-2963]: https://pins-ds.atlassian.net/browse/THEODW-2963?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


odw/test/integration_test/etl/curated/test_appeal_document_curated_process.py
```diff
@@ -0,0 +1,464 @@
...
+        assert_dataframes_equal(expected_df, actual_df)
+
+    def write_source_table(self, spark, table_df: DataFrame):
+        self.write_existing_table(
+            spark,
+            table_df,
+            "appeal_document",
+            "odw_harmonised_db",
+            "odw-harmonised",
+            "appeal_document",
```
HarrisonBoyleThomas: You might need to add variation to the table name/blob name, since if you run in parallel you might get conflicts

odw/test/integration_test/etl/curated/test_appeal_document_curated_process.py
```diff
@@ -0,0 +1,464 @@
...
+            "overwrite",
+        )
+
+    def write_target_table(self, spark, table_df: DataFrame):
+        self.write_existing_table(
+            spark,
+            table_df,
+            "appeal_document",
```
HarrisonBoyleThomas, line 195: Same here

odw/test/integration_test/etl/curated/test_appeal_document_curated_process.py
```diff
@@ -0,0 +1,464 @@
...
+            "overwrite",
+        )
+
+    def write_empty_target_table(self, spark):
+        empty_target_df = spark.createDataFrame([], _curated_schema())
+        self.write_target_table(spark, empty_target_df)
+
+    def test__appeal_document_curated_process__run__initial_load_matches_legacy(
+        self,
+    ):
```
HarrisonBoyleThomas, line 208: Would be good to add a gherkin syntax docstring to explain the test case

odw/test/integration_test/etl/curated/test_appeal_document_curated_process.py
```diff
@@ -0,0 +1,464 @@
...
+from odw.test.integration_test.etl.etl_test_case import ETLTestCase
+from odw.test.util.session_util import PytestSparkSessionUtil
+
+pytestmark = pytest.mark.xfail(reason="Curated logic not implemented yet")
+
+
+def _harmonised_schema():
+    return StructType(
+        [
+            StructField("AppealsDocumentMetadataID", LongType(), True),
```
HarrisonBoyleThomas: The real data seems to be `IntegerType`, but this should be fine i think

odw/test/unit_test/etl/curated/test_appeal_document_curated_process.py
```diff
@@ -0,0 +1,366 @@
...
+            data_to_write, _ = inst.process(source_data=source_data)
+
+        df = data_to_write[inst.OUTPUT_TABLE]["data"]
+
+        rows = {row["documentId"]: row["caseType"] for row in df.select("documentId", "caseType").collect()}
+
+        assert rows["doc-1"] == "D"
+        assert rows["doc-2"] == "W"
+        assert rows["doc-3"] == "X"
+
```
HarrisonBoyleThomas: I think it would be good to flatten these into a single assertion. i.e. you could just create the expected dataframe with DWX rows

odw/test/unit_test/etl/curated/test_appeal_document_curated_process.py
```diff
@@ -0,0 +1,366 @@
...
+            _harmonised_row(documentId="doc-2", caseType="Planning Appeal (A)", IsActive="Y"),
+            _harmonised_row(documentId="doc-3", caseType="Lawful Development Certificate Appeal", IsActive="Y"),
+        ]
+
+        source_data = {
+            "source_data": spark.createDataFrame(source_rows, schema=_harmonised_schema()),
+            "target_exists": False,
+        }
+
+        with mock.patch("odw.core.etl.transformation.curated.appeal_document_curated_process.LoggingUtil"):
```
HarrisonBoyleThomas: You might need to patch `Util.get_storage_account`

HarrisonBoyleThomas: This will be the case for most of the unit tests, and also possibly the int test

Fred83200: Added

odw/test/unit_test/etl/curated/test_appeal_document_curated_process.py
Fred83200: Done - Did it via a DF assertion

odw/test/integration_test/etl/curated/test_appeal_document_curated_process.py
Fred83200: Good catch - Changed it

odw/test/integration_test/etl/curated/test_appeal_document_curated_process.py
Fred83200: Added to each tests

odw/test/integration_test/etl/curated/test_appeal_document_curated_process.py
Fred83200: added the uuid import and also a setup class to create a unique path suffix so they don't clash if run in parallel

odw/test/integration_test/etl/curated/test_appeal_document_curated_process.py
Fred83200: added the uuid import and also a setup class to create a unique path suffix so they don't clash if run in parallel

HarrisonBoyleThomas (approved):

PR #2469, opened 2026-04-20
Fix/theodw 3006 irregular col name

HarrisonBoyleThomas:
Ticket: https://pins-ds.atlassian.net/browse/THEODW-3006

Fixes an issue caused by a metadata column added in the source data called `@odata.context`, which is not needed


stef-solirius (approved):
LGTM, thanks

PR #2478, opened 2026-04-23
Feat/3059 convert ut cappeal event curated mipins

KranthiRayipudi:
[https://pins-ds.atlassian.net/browse/THEODW-3049](url)

Updated timestamp handling to convert UTC timestamps to Europe/London time, including Daylight Saving Time (DST).
Applied the UTC‑to‑London conversion consistently across all relevant timestamp fields:
eventStartDateTime
eventEndDateTime
IngestionDate
ValidTo
Ensured all converted columns are stored as proper timestamp types.
Updated the write logic to overwrite data in the existing curated table location instead of dropping and recreating the table.
Retrieved the table’s physical storage path dynamically using DESCRIBE EXTENDED and wrote the data back to the same location using overwrite mode.
This maintains table metadata while safely refreshing the curated data with the updated timestamp logic.

image

### DIFF
--- a/workspace/notebook/appeal_event_curated_mipins.json
+++ b/workspace/notebook/appeal_event_curated_mipins.json
@@ -20,7 +20,7 @@
 				"spark.dynamicAllocation.enabled": "false",
 				"spark.dynamicAllocation.minExecutors": "2",
 				"spark.dynamicAllocation.maxExecutors": "2",
-				"spark.autotune.trackingId": "21e89f80-95a3-4698-8456-3d38082ba38e"
+				"spark.autotune.trackingId": "fd1d3712-6ba4-4400-9a1f-25ca2e98cbab"
 			}
 		},
 		"metadata": {
@@ -58,7 +58,7 @@
 					"from datetime import datetime, date\n",
 					"import pyspark.sql.functions as F"
 				],
-				"execution_count": null
+				"execution_count": 49
 			},
 			{
 				"cell_type": "code",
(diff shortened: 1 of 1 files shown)


workspace/notebook/appeal_event_curated_mipins.json
```diff
@@ -110,33 +119,33 @@
...
-					"        Migrated,\n",
-					"        ODTSourceSystem,\n",
-					"        SourceSystemID,\n",
-					"        IngestionDate,\n",
-					"        ValidTo,\n",
-					"        RowID,\n",
-					"        IsActive\n",
-					"    FROM vw_Appeal_event_curated_mipins\n",
-					"\"\"\")\n",
+					"view_df = spark.sql(\"SELECT * FROM odw_curated_db.vw_Appeal_event_curated_mipins\")"
```
HarrisonBoyleThomas, line 122: Is `*` equivalent to the column list in the original query?

workspace/notebook/appeal_event_curated_mipins.json
```diff
@@ -110,33 +119,33 @@
...
+					"'eventStartDateTime',    \n",
+					"'eventEndDateTime', \n",
+					"'IngestionDate',\n",
+					"'ValidTo',\n",
+					"]\n",
 					"\n",
+					"# Apply the UTC to London conversion for each column\n",
+					"final_view_df = view_df\n",
+					"for column_name in timestamp_columns:\n",
+					"    final_view_df = final_view_df.withColumn(column_name, F.to_timestamp(F.from_utc_timestamp(F.col(column_name), 'Europe/London'), \"yyyy-MM-dd'T'HH:mm:ss.SSS\"))"
```
HarrisonBoyleThomas: Might be better to use https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.withColumns.html here to boost performance

workspace/notebook/appeal_event_curated_mipins.json
KranthiRayipudi: yes

workspace/notebook/appeal_event_curated_mipins.json
KranthiRayipudi: I have cloned the same implementation that was previously done for other entities by someone else in our team. If I change this logic now, I would need to apply the same changes across all related entities, which would take additional time.

HarrisonBoyleThomas: You could just change this entity for now, other entities could be improved as part of a separate piece of work

KranthiRayipudi: I have implemented the suggested changes. Could you please check and confirm?

HarrisonBoyleThomas (approved):

PR #2479, opened 2026-04-23
feat: strengthened anonymisation with postcode, address and improved diagnostics

stef-solirius:
https://pins-ds.atlassian.net/browse/THEODW-3009

## Summary

- Adds `PostcodeStrategy` (keeps outward code only, e.g. `SW1W 9SP` → `SW1W`) mapped to the Purview `UK Postcode` classification
- Adds `AddressStrategy` to redact flat and struct address fields (`MICROSOFT.PERSONAL.PHYSICALADDRESS`, `All Physical Addresses`); struct fields preserve the postcode outward code
- Aligns `NameMaskStrategy` to return `REDACTED` (previously kept first/last char)
- Adds `py_purview_create_postcode_classification` notebook to register the `UK Postcode` custom classification rule in Purview
- Adds `combined_odw` notebook (flattened package entry point for Synapse)
- Improves anonymisation engine diagnostics: `purview.columns_raw` logs every column and its raw Purview classifications; `apply_from_purview.columns_skipped` logs each rejected column with the exact rejection reason (`not_in_dataframe` / `no_allowlisted_classification` / `no_matching_strategy`)

## Test plan

- [ ] Run unit tests: `poetry run pytest tests/unit_test/anonymisation/ -v`
- [ ] Verify `PostcodeStrategy` keeps outward code and redacts inward code for both full and null postcodes
- [ ] Verify `AddressStrategy` REDACTs flat string address columns and struct address fields while preserving postcode outward code in struct
- [ ] Re-run `py_etl_orchestrator` for `service-user` in DEV and confirm `apply_from_purview.columns_selected` includes `postcode` and `changed_rows > 0`
- [ ] Confirm `purview.columns_skipped` shows only `PotentialID`/`PotentialDate` columns as expected non-PII


odw/core/anonymisation/base.py
```diff
@@ -188,12 +140,60 @@ def apply(self, df: DataFrame, column: str, seed: Column, context: dict) -> Data
         return df.withColumn(column, BaseStrategy.random_int_from_seed(seed, 20000, 100000).cast("int"))
 
 
+class PostcodeStrategy(BaseStrategy):
+    classification_names = {"UK Postcode"}
+
+    def apply(self, df: DataFrame, column: str, seed: Column, context: dict) -> DataFrame:
+        col_expr = F.col(column).cast("string")
+        result = F.when(F.col(column).isNull(), None).otherwise(F.split(col_expr, " ").getItem(0))
```
Fred83200: Just thinking here the handling is a bit weak except if that's intended. 
This F.split will fail if you pass in a double spaced postcode or leading space, no space etc.... 
I'd change it to trim and split on whitespace or even use a regex extracting

odw/core/anonymisation/base.py
```diff
@@ -188,12 +140,60 @@ def apply(self, df: DataFrame, column: str, seed: Column, context: dict) -> Data
...
+                for field in schema.fields:
+                    if field.name.lower() == "postcode":
+                        pc = F.col(f"{column}.{field.name}")
+                        redacted_fields.append(F.when(pc.isNull(), None).otherwise(F.split(pc.cast("string"), " ").getItem(0)).alias(field.name))
+                    else:
+                        redacted_fields.append(F.lit("REDACTED").alias(field.name))
+                redacted_struct = F.struct(*redacted_fields)
+                return df.withColumn(column, F.when(F.col(column).isNotNull(), redacted_struct).otherwise(None))
+
+        return df.withColumn(column, F.when(F.col(column).isNotNull(), F.lit("REDACTED")).otherwise(None))
```
Fred83200: Here I'm thinking that for a simple fields null stays null but if it's a struct field a nested null will become "REDACTED"
Not sure if that's intentional ?

odw/core/anonymisation/base.py
```diff
@@ -188,12 +140,60 @@ def apply(self, df: DataFrame, column: str, seed: Column, context: dict) -> Data
...
+
+        col_type = dict(df.dtypes).get(column)
+
+        if col_type and col_type.startswith("struct"):
+            schema = df.schema[column].dataType
+            if isinstance(schema, StructType):
+                redacted_fields = []
+                for field in schema.fields:
+                    if field.name.lower() == "postcode":
+                        pc = F.col(f"{column}.{field.name}")
```
Fred83200: This I think is a bit risky - if the top lvl col name or nested field contains dots/special char, Spark can misinterpret it
Might be safe to use backticks or an helper just an idea though

odw/core/anonymisation/engine.py
```diff
@@ -529,26 +529,46 @@ def _apply_from_purview_explicit(
...
             if not actual_col:
+                skipped.append({"column": col_name, "reason": "not_in_dataframe", "classifications": item.get("classifications")})
                 continue
-            classes = set(item.get("classifications") or [])
-            if allowlist is not None:
-                classes = {c for c in classes if c in allowlist}
+            raw_classes = set(item.get("classifications") or [])
+            classes = {c for c in raw_classes if c in allowlist} if allowlist is not None else raw_classes
             if not classes:
+                skipped.append({"column": col_name, "reason": "no_allowlisted_classification", "raw_classifications": sorted(raw_classes)})
```
Fred83200, line 559: Just to make sure : Are these present in config.classification_allowlist if not they will be filtered out here and logges as "no_allowlisted_classification" meaning your new strategie will never be applied

odw/core/anonymisation/engine.py
```diff
@@ -529,26 +529,46 @@ def _apply_from_purview_explicit(
...
                 continue
             if any(classes.intersection(strat.classification_names) for strat in self.strategies):
                 target_cols.append(actual_col)
```
Fred83200: target_cols is currently a list, so if Purview returns the same column multiple times (which can happen depending on scan results) the same column might be anonymised multiple times
This is usually harmless for constant strategies (e.g. REDACTED) but could cause issues for non idempotent strategies ( randomised or hash based ones)

maybe deduplicating while preserving order like using an ordered set pattern or checking before appending

Might not be the case - Just double checking

odw/test/unit_test/anonymisation/test_unit_anonymisation_engine.py
```diff
@@ -397,9 +396,251 @@ def test__horizon_standardisation__anonymisation_handles_null_values(self):
...
         # First row should be anonymised
-        assert rows[0]["First Name"] == "J***"
+        assert rows[0]["First Name"] == "REDACTED"
         assert rows[0]["Email Address"] == "836f82db99121b3481011f16b49dfa5fbc714a0d1b1b9f784a1ebbbf5b39577f"
 
         # Second row nulls should remain null
         assert rows[1]["First Name"] is None
         assert rows[1]["Email Address"] is None
+
+    def test__anonymisation__postcode_keeps_first_block(self):
```
Fred83200, line 406: This test is good was thinking maybe you want to add some edge cases to make it more robust like : 
Postcode with multiple spaces
leading/traling spaces
lower case
Postcode with no space

Maybe adding a test where the Purview classification is present but not allowlisted (should be skipped)
or a test where the same column appears multiple times in Purview results (to validate deduplication behaviour)

odw/core/anonymisation/base.py
stef-solirius: You're right, I have trimmed first and splitted on \s+ so Spark's split accepts as a regex

odw/core/anonymisation/base.py
stef-solirius: Yeah, not what we want. Changed it so Null nested fields now stay null

odw/core/anonymisation/base.py
stef-solirius: Good point, fixed that too

odw/core/anonymisation/engine.py
stef-solirius: It's okay for the mock_policy to be oiutdates, but I'll check in odw-config/anonymisation just to be sure

odw/core/anonymisation/engine.py
stef-solirius: Added a fix

odw/test/unit_test/anonymisation/test_unit_anonymisation_engine.py
stef-solirius: Thanks, fixed that test to explicitly pass an allowlist

Fred83200 (dismissed):

Fred83200 (approved):

PR #2480, opened 2026-04-23
Reducing pytest parallelism + optimising Spark test config

Fred83200:
[THEODW-3044](https://pins-ds.atlassian.net/browse/THEODW-3044)
What this PR does : 

- Reduce unit test parallelism from -n 4 to -n 2
- Run integration tests with -n 1 to avoid multiple concurrent Spark sessions
- Add --dist loadscope and --maxfail=1 to improve test execution behaviour
- Tune PySpark test configuration to reduce memory pressure and shuffle overhead
- Optimise test Spark session setup in session_util.py

[THEODW-3044]: https://pins-ds.atlassian.net/browse/THEODW-3044?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/odw/test/util/session_util.py
+++ b/odw/test/util/session_util.py
@@ -10,33 +10,43 @@ import shutil
 
 class PytestSparkSessionUtil(metaclass=Singleton):
     """
-    Class to manage the testing session itself. This initiallises a single spark context per worker thread,
-    alongside a separate DB/filesystem per thread. This class is a singleton, so only one instance can ever
-    be created
+    Class to manage the testing session itself. This initialises a single Spark session per
+    worker thread, alongside a separate DB/filesystem per thread. This class is a singleton,
+    so only one instance can ever be created.
 
-    # Usage
-    - `PytestSessionUtil().get_spark_session()`
-    - `PytestSessionUtil().get_thread_id()`
-    - `PytestSessionUtil().get_spark_warehouse_name()`
+    Usage
+    - `PytestSparkSessionUtil().get_spark_session()`
+    - `PytestSparkSessionUtil().get_thread_id()`
+    - `PytestSparkSessionUtil().get_spark_warehouse_name()`
     """
 
     DATABASE_NAMES = ["odw_standardised_db", "odw_harmonised_db", "odw_curated_db"]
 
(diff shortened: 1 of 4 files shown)


HarrisonBoyleThomas:
I would drop the `maxfail` argument since this seems to stop all tests if just 1 fails

odw/test/util/session_util.py
```diff
@@ -10,33 +10,43 @@
...
+            .appName(f"pytest-spark-{self.get_thread_id()}")
+            .config("spark.sql.extensions", "io.delta.sql.DeltaSparkSessionExtension")
             .config("spark.sql.catalog.spark_catalog", "org.apache.spark.sql.delta.catalog.DeltaCatalog")
             .config("spark.sql.warehouse.dir", self.get_spark_warehouse_name())
             .config("spark.ui.enabled", False)
-        ).getOrCreate()
+            .config("spark.driver.memory", "1g")
+            .config("spark.sql.shuffle.partitions", "4")
+            .config("spark.default.parallelism", "2")
+            .config("spark.sql.adaptive.enabled", "false")
```
HarrisonBoyleThomas, line 39: How come you disabled this? I was thinking this would boost runtime performance

Fred83200: Yes but in this case I disabled it for stability and memory control rather than runtime speed. In our CI we're running multpiple Spark sessions in parallel, AQE can increase the memory usage unpredictably which I think been contributing to the pressure on the Py4J crashes I saw. 
That's my thought

Fred83200: If later we reduce the parallelism more or increase the agent memory then we can def re enable it 
it's a trade between being less optimal in exec but more predictable and lower memory usage accross the tests than runs in parallel

HarrisonBoyleThomas: Sounds good

HarrisonBoyleThomas (dismissed):

HarrisonBoyleThomas (approved):

PR #2485, opened 2026-04-27
Feat/theodw 2921 restructure appeal document notebooks

priyanka-challa:
**PR Template**

JIRA Ticket Reference  :
    [Restructure Appeal document Notebooks py_sb_horizon_harmonised_appeal_document and appeal_document](https://pins-ds.atlassian.net/browse/THEODW-2921)

 Summary of the work : 
-  `appeal_document_harmonisation_process.py `
-  `appeal_document_curated_process.py`

Moved Appeal Document harmonised logic into odw core as a dedicated `AppealDocumentHarmonisationProcess` class
Moved Appeal Document curated logic into odw core as a dedicated `AppealDocumentCuratedProcess` class

Changes
- `odw/core/etl/etl_process_factory.py` — registered new Appeal Document harmonisation and curated processes
- `odw/curated/appeal_document_curated_process.py` — new curated process 
- `odw/harmonised/appeal_document_harmonisation_process.py` — new harmonisation process 
- remove the xfail from appeal document notebook tests as logic is implemented
- Rename entity name "appeal_document_harmonised_process" →  "appeal_document_harmonisation_process"
- Add TEMP_PK field to service bus schema and row builder to match
 _load_service_bus_data which computes MD5(documentId, filename, version)
- Add new service bus fields: caseNumber, caseworkTypeGroup,
 caseworkTypeAbbreviation, versionFilename, incomingOutgoingExternal,
 publishedStatus
- Add sb_primary_keys to source_data dict required by process
- Mock Util.get_storage_account alongside LoggingUtil to prevent
 real storage account lookups during tests
- Add setup_method to create odw_harmonised_db.appeal_document table
 so DESCRIBE EXTENDED succeeds during process execution
- Integration: add _ensure_output_table helper to create Delta table
- Rename source_data key "source_data" → "harmonised_appeal_docs" to match updated curated process input expectation
- Comment out inactive document test row as there is no filter logic applied in process method
- Rename setup() → setup_method() for pytest compatibility
- Mock notebookutils in conftest.py for local test execution 
- New appeal document ETL tests import production code that requires notebookutils (Synapse-only). Importing the mock first prevents ModuleNotFoundError during test collection.

**Appeal Document Test Results:**
image


odw/core/etl/transformation/harmonised/appeal_document_harmonisation_process.py
```diff
@@ -1,25 +1,457 @@
...
+        # ---------------------------------------------------------
+        # Step 7: Compute RowID from the final_df.
+        # ---------------------------------------------------------
+
+        row_id_expr = F.md5(F.concat(*[F.coalesce(F.col(c).cast("string"), F.lit(".")) for c in _APPEAL_DOCUMENT_ROW_ID_COLUMNS]))
+        final_df = final_df.withColumn("RowID", row_id_expr)
+
+        insert_count = final_df.count()
+
+        # ---------------------------------------------------------
```
stef-solirius: Don't think that the NSIP equivalent includes this Step, it just lets the write figure the path out. You can see the cost of leaving it in: every unit test now needs a setup_method that stands up a Hive DB and a dummy Delta table, and  the integration tests added _ensure_output_table() just to keep this process happy. Please move the path resolution into the writer (or into load_data if it really has to live here), then drop the table-creation crutches from the tests.

odw/core/etl/transformation/harmonised/appeal_document_harmonisation_process.py
```diff
@@ -1,25 +1,457 @@
...
+        }
+
+    def _load_service_bus_data(self) -> DataFrame:
+        """
+        Get data out of the service bus with additional fields needed for Horizon data.
+        Computes a primary key: MD5(CONCAT(documentId, filename, version))
+        """
+        return self.spark.sql(f"""
+            SELECT DISTINCT
+                MD5(CONCAT(documentId, filename, version)) AS {self.PRIMARY_KEY}
```
stef-solirius: This MD5 primary key can silently come out as NULL. If any of the three is NULL, CONCAT returns NULL, MD5 returns NULL, and every NULL-PK row collides into the same "key". The Horizon side already wraps each input in COALESCE. Pplease mirror that here, or assert these columns are NOT NULL upstream and let it fail loudly if they aren't.

odw/core/etl/transformation/harmonised/appeal_document_harmonisation_process.py
```diff
@@ -1,25 +1,457 @@
...
+        """)
+
+    def _load_service_bus_primary_keys(self) -> DataFrame:
+        """
+        Load the distinct set of primary keys from the service-bus table.
+        Used to derive the Migrated flag (1 if the PK exists in SB, else 0).
+        """
+        return self.spark.sql(f"""
+            SELECT DISTINCT
+                MD5(CONCAT(documentId, filename, version)) AS {self.PRIMARY_KEY}
```
stef-solirius: Same concern as above, the MD5 primary key can still silently come out as NULL

odw/core/etl/transformation/harmonised/appeal_document_harmonisation_process.py
stef-solirius: Actually, I discussed with Fred and Harrison, it's definitely better to leave all the queries to load_data. process shouldn't have any data loaded, just transformation.

odw/test/integration_test/etl/harmonised/test_appeal_document_harmonisation_process.py
```diff
@@ -311,31 +342,38 @@ def test__appeal_document_harmonisation_process__process__uses_latest_horizon_sn
...
+        _ensure_output_table(spark)
 
+        sb_df = spark.createDataFrame([], _service_bus_schema())
         source_data = {
-            "service_bus_data": spark.createDataFrame([], _service_bus_schema()),
+            "service_bus_data": sb_df,
             "horizon_data": spark.createDataFrame(
                 [
-                    _horizon_row(documentId="doc-old", ingested_datetime="2025-01-10T12:00:00"),
+                    # _horizon_row(documentId="doc-old", ingested_datetime="2025-01-10T12:00:00"),
```
stef-solirius: Not sure why the older row was commented out rather than deleted, and the snapshot filter (WHERE ingested_datetime = MAX(...)) lives inside _load_horizon_data — which the test bypasses by handing process() a pre-built DataFrame. So right now nothing in the test suite actually proves the latest-snapshot filter works. Probs worth deleting the test then.

odw/test/unit_test/etl/curated/test_appeal_document_curated_process.py
```diff
@@ -149,11 +146,11 @@ def test__appeal_document_curated_process__process__filters_to_active_rows_only(
 
         source_rows = [
             _harmonised_row(documentId="doc-active", IsActive="Y"),
-            _harmonised_row(documentId="doc-inactive", IsActive="N"),
+            # _harmonised_row(documentId="doc-inactive", IsActive="N"),
```
stef-solirius: Same here as for the test_apepal_document_harmonised_process.py - worth removing this too.

odw/core/etl/transformation/harmonised/appeal_document_harmonisation_process.py
```diff
@@ -1,25 +1,457 @@
...
+            )
+            .distinct()
+        )
+
+        # ---------------------------------------------------------
+        # Step 2: Align Horizon columns to SB columns and union
+        # ---------------------------------------------------------
+        LoggingUtil().log_info(f"Combining data for {self.OUTPUT_TABLE}")
+        horizon_joined = horizon_joined.select(service_bus_data.columns)
+        combined = service_bus_data.union(horizon_joined)
```
stef-solirius: union is positional! it lines columns up by index, not by type. The Horizon path passes Doc.virusCheckStatus, Doc.size, etc. through with  no explicit cast, while the SB schema has integer types in places (virusCheckStatus, caseId, version). In production, we either need the Horizon side casts to match SB exactly, or we need this to be UnionByName with a schema check so a future drift fails loudly instead of silently coercing. Please confirm the production schemas agree and add an assertion either way.

odw/core/etl/transformation/harmonised/appeal_document_harmonisation_process.py
priyanka-challa: I have moved path resolution to load_data as suggested and removed the setup_method and _ensure_output_table from the tests as it no longer required

odw/core/etl/transformation/harmonised/appeal_document_harmonisation_process.py
priyanka-challa: I have added the COALESCE for all the 3 columns and it should handle the NULL check correctly

odw/core/etl/transformation/harmonised/appeal_document_harmonisation_process.py
priyanka-challa: updated the same here as suggested above

odw/test/integration_test/etl/harmonised/test_appeal_document_harmonisation_process.py
priyanka-challa: I have removed the test as suggested and  it's not doing the purpose

odw/test/unit_test/etl/curated/test_appeal_document_curated_process.py
priyanka-challa: I have removed the test here as well as suggested and it's not doing the purpose

odw/core/etl/transformation/harmonised/appeal_document_harmonisation_process.py
priyanka-challa: I have updated to use unionByName as suggested

odw/core/etl/transformation/harmonised/appeal_document_harmonisation_process.py
```diff
@@ -1,25 +1,460 @@
...
+            "horizon_data": horizon_data,
+            "aie_data": aie_data,
+            "sb_primary_keys": sb_primary_keys,
+            "table_path": table_path,
+        }
+
+    def _load_service_bus_data(self) -> DataFrame:
+        """
+        Get data out of the service bus with additional fields needed for Horizon data.
+        Computes a primary key: MD5(CONCAT(documentId, filename, version))
```
stef-solirius: Very sorry one more thing, in the synapse notebook py_sb_horizon_harmonised_appeal_document, we create the primary key by hashing MD5(CONCAT(documentId, filename, version, documentURI)) AS {primary_key}

...what happened to documentURI here?

odw/core/etl/transformation/harmonised/appeal_document_harmonisation_process.py
```diff
@@ -1,25 +1,460 @@
...
+        """)
+
+    def _load_service_bus_primary_keys(self) -> DataFrame:
+        """
+        Load the distinct set of primary keys from the service-bus table.
+        Used to derive the Migrated flag (1 if the PK exists in SB, else 0).
+        """
+        return self.spark.sql(f"""
+            SELECT DISTINCT
+                MD5(CONCAT(COALESCE(documentId, ''), COALESCE(filename, ''), COALESCE(CAST(version AS STRING), ''))) AS {self.PRIMARY_KEY}
```
stef-solirius, line 239: also here, what happened to documentURI?

odw/core/etl/transformation/harmonised/appeal_document_harmonisation_process.py
```diff
@@ -1,25 +1,460 @@
...
+
+        # ---------------------------------------------------------
+        # Step 5: Derive Migrated flag (1 if PK exists in SB table, else 0)
+        # ---------------------------------------------------------
+        sb_keys = sb_primary_keys.withColumnRenamed(pk, "sb_pk")
+        calcs = (
+            calcs.join(sb_keys, calcs[pk] == sb_keys["sb_pk"], "left_outer")
+            .withColumn("Migrated", F.when(F.col("sb_pk").isNotNull(), "1").otherwise("0"))
+            .drop("sb_pk")
+        )
```
stef-solirius, line 387: Just to confirm , the change in Migrated-flag values (always-0 → 1/0) is intended? since it changes downstream view of the data.

In the synapse notebook the Raw subquery key (MD5(CONCAT(documentId, filename, version))) never matches CurrentRow.{primary_key} (which includes documentURI), so Migrated effectively always resolves to '0'. In the Python it correctly evaluates to '1' for SB records and '0' for Horizon records. Looks like you are fixing the but, but it might be a real change in output values...

odw/core/etl/transformation/harmonised/appeal_document_harmonisation_process.py
priyanka-challa: Hi @stef-solirius Initially I have added documentURI but later Fred checked and suggested the Legacy Notebook is not using the documentURI so I have changed it to match with Legacy Notebook

https://github.com/Planning-Inspectorate/odw-synapse-workspace/pull/2463#discussion_r3110488834

odw/core/etl/transformation/harmonised/appeal_document_harmonisation_process.py
priyanka-challa: here also same matching with Legacy Notebook behaviour

Fred83200: Legacy does use documentURI when creating the main primary key for both Service Bus and Horizon:
MD5(CONCAT(documentId, filename, version, documentURI)) AS TEMP_PK
This piece is in the Legacy notebook. 

So those  2 need to include documentURI : 
_load_service_bus_data()
_load_service_bus_primary_keys()

odw/core/etl/transformation/harmonised/appeal_document_harmonisation_process.py
Fred83200: Legacy does use documentURI when creating the main primary key for both Service Bus and Horizon:
MD5(CONCAT(documentId, filename, version, documentURI)) AS TEMP_PK
This piece is in the Legacy notebook. 

So those  2 need to include documentURI : 
_load_service_bus_data()
_load_service_bus_primary_keys()

odw/core/etl/transformation/harmonised/appeal_document_harmonisation_process.py
Fred83200: Main record uses : 
MD5(CONCAT(documentId, filename, version, documentURI))
 
The raw migrated join uses : 
 
MD5(CONCAT(documentId, filename, version))
 
So it will never match and migrated will stay "0"
This is not matching Legacy behaviour notebook

odw/core/etl/transformation/harmonised/appeal_document_harmonisation_process.py
priyanka-challa: As discussed I have modified only _load_service_bus_data with documentURI to generate the primary key

odw/test/unit_test/etl/harmonised/test_appeal_document_harmonisation_process.py
```diff
@@ -181,6 +191,9 @@ def _service_bus_row(**overrides):
         "IsActive": "Y",
     }
     row.update(overrides)
+    # Compute TEMP_PK to match _load_service_bus_data: MD5(CONCAT(documentId, filename, version))
+    if "TEMP_PK" not in row:
+        row["TEMP_PK"] = hashlib.md5(f"{row['documentId']}{row['filename']}{row['version']}".encode()).hexdigest()
```
stef-solirius: This hashing is still missing documentURL

odw/test/integration_test/etl/harmonised/test_appeal_document_harmonisation_process.py
```diff
@@ -181,6 +191,9 @@ def _service_bus_row(**overrides):
         "IsActive": "Y",
     }
     row.update(overrides)
+    # Compute TEMP_PK to match _load_service_bus_data: MD5(CONCAT(documentId, filename, version))
+    if "TEMP_PK" not in row:
+        row["TEMP_PK"] = hashlib.md5(f"{row['documentId']}{row['filename']}{row['version']}".encode("utf-8")).hexdigest()
```
stef-solirius: This is also missing documenURL

stef-solirius (approved):
LGTM, thanks Priyanka

PR #2496, opened 2026-05-05
Feat/theodw 2989 dart api tests

HarrisonBoyleThomas:
Ticket: https://pins-ds.atlassian.net/browse/THEODW-2989

Add unit and integration tests for the dart api curation refactor


odw/core/etl/transformation/curated/dart_api_curation_process.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class DartAPICurationProcess(CurationProcess):
+    HARMONISED_APPEAL_HAS = "appeal_has"
+    HARMONISED_PINS_LPA = "pins_lpa"
+    HARMONISED_SB_SERVICE_USER = "sb_service_user"
+    HARMONISED_SB_APPEAL_EVENT = "sb_appeal_event"
+    HARMONISED_PINS_INSPECTOR = "horizon_pins_inspector"
+    HARMONISED_ENTRAID = "entraid"
+    HARMONISED_APPEAL_S78 = "horizon_appeal_s78"
```
stef-solirius: Are you sure this this is the right harmonised table? Looks like the notebook queries odw_harmonised_db.appeal_s78, so the union of Horizon and SB

odw/test/integration_test/etl/curated/test_dart_api_curation_process.py
```diff
@@ -0,0 +1,1511 @@
...
+                    caseId=7, lpaCode="G", caseReference=60000007, inspectorId="insId7", caseOfficerId="insId1"
+                ),  # For the dropped appeal_event row
+                generate_harmonised_horizon_appeal_s78_row(
+                    caseId=8, lpaCode="H", caseReference=60000008, inspectorId="insId10", caseOfficerId="insId7"
+                ),  # For the dropped entraid_ins row
+                generate_harmonised_horizon_appeal_s78_row(
+                    caseId=9, lpaCode="I", caseReference=60000009, inspectorId="insId7", caseOfficerId="insId10"
+                ),  # For the dropped entraid_co row
+            ),
+            schema=generate_harmonised_appeal_has_schema(),
```
stef-solirius: I might be wrong, but shouldn't this be        
      schema=generate_harmonised_horizon_appeal_s78_schema()?
The rows were generated with generate_harmonised_horizon_appeal_s78_row, otherwise we will have a schema mismatch when the data is eventually read back from the written table.

odw/test/integration_test/etl/curated/test_dart_api_curation_process.py
```diff
@@ -0,0 +1,1511 @@
...
+            ),
+            schema=generate_cleaned_schema(),
+        )
+        expected_curated_table = f"{test_case}_dart_api"
+        with (
+            mock.patch.object(DartAPICurationProcess, "__init__", return_value=None),
+            mock.patch.object(DartAPICurationProcess, "HARMONISED_APPEAL_HAS", appeal_has_table),
+            mock.patch.object(DartAPICurationProcess, "HARMONISED_PINS_LPA", pins_lpa_table),
+            mock.patch.object(DartAPICurationProcess, "HARMONISED_SB_SERVICE_USER", sb_service_user_table),
+            mock.patch.object(DartAPICurationProcess, "HARMONISED_SB_APPEAL_EVENT", sb_appeal_event),
```
stef-solirius: HARMONISED_SB_APPEAL_EVENT and HARMONISED_PINS_INSPECTOR are string table-name constants. Patching them with DataFrames means the implementation would receive a     
  DataFrame object where it expects a string and would attempt to call spark.table(<DataFrame>), raising a TypeError at runtime. I think we should match the pattern used by every other mock in the same blockm like:
                                                                                              
  mock.patch.object(DartAPICurationProcess, "HARMONISED_SB_APPEAL_EVENT", sb_appeal_event_table),                                                                      
  mock.patch.object(DartAPICurationProcess, "HARMONISED_PINS_INSPECTOR", horizon_pins_inspector_table),

odw/test/integration_test/etl/curated/test_dart_api_curation_process.py
```diff
@@ -0,0 +1,1511 @@
...
+                    caseReference="60000004",
+                    lpaCode="D",
+                ),
+                generate_cleaned_row(
+                    caseId=5,
+                    caseReference="60000005",
+                    lpaCode="E",
+                ),
+            ),
+            schema=generate_cleaned_schema(),
```
stef-solirius, line 1443: Missing from generate_cleaned_row base dict:                                                                                                                         
  "lpaName": None,
  "appellantName": None,                                                                                                                                               
  "typeOfEvent": None,       
  "startDateOfTheEvent": None,
  "inspectorName": None,      
  "caseOfficerName": None,
  "inspectorQualifications": None,

Otherwise they will default to null and the assert_dataframes_equal comparison will fail because expected has null for lpaName etc., but actual has some real values.

odw/test/integration_test/etl/curated/test_dart_api_curation_process.py
```diff
@@ -0,0 +1,1511 @@
...
+
+
+class TestDartAPICurationProcess(ETLTestCase):
+    def assert_etl_process(self, test_case: str):
+        spark = PytestSparkSessionUtil().get_spark_session()
+        appeal_has = spark.createDataFrame(
+            (
+                generate_harmonised_appeal_has_row(caseId=1, lpaCode="A", caseReference=60000001, inspectorId="insId1", caseOfficerId="insId7"),
+                generate_harmonised_appeal_has_row(
+                    caseId=2, lpaCode="B", caseReference=60000002, inspectorId="insId2", inspcaseOfficerIdectorId="insId6"
```
stef-solirius: inspcaseOfficerIdectorId is not a valid ke,  it's a garbled form of caseOfficerId. The base overrides dict merge silently discards it, so caseOfficerId stays as  "" (the default). If this is intentional (to test null case officer), rename it to something like # caseOfficerId intentionally omitted to test null case officer. If not, fix it to caseOfficerId="insId6".

odw/test/unit_test/etl/curated/test_dart_api_curation_process.py
```diff
@@ -0,0 +1,2327 @@
...
+            + [
+                generate_harmonised_horizon_appeal_s78_row(caseId=4, IsActive="N"),  # Inactive row
+                generate_harmonised_horizon_appeal_s78_row(caseId=5, lpaCode="Q9999"),
+                generate_harmonised_horizon_appeal_s78_row(caseId=6, caseReference=1),
+                generate_harmonised_horizon_appeal_s78_row(caseId=6, ODTSourceSystem="Some other system"),
+            ],
+            schema=generate_harmonised_horizon_appeal_s78_schema(),
+        )
+        self.write_existing_table(spark, appeal_has_data, test_case, "odw_harmonised_db", "odw-harmonised", test_case, "overwrite")
+        expected_data = spark.createDataFrame(kept_rows, schema=generate_harmonised_appeal_has_schema())
```
stef-solirius: I think this should also be expected_data = spark.createDataFrame(kept_rows, schema=generate_harmonised_horizon_appeal_s78_schema()) not has.

odw/test/unit_test/etl/curated/test_dart_api_curation_process.py
```diff
@@ -0,0 +1,2327 @@
...
+                generate_aggregate_row(
+                    caseId=2,
+                    caseReference="2",
+                    lpaCode="B",
+                    applicationReference="",  # Should be replaced with UNKNOWN
+                    enforcementNotice=False,  # Should become True
+                    isCorrectAppealType=False,  # Should become True
+                    appellantCostsAppliedFor=False,  # Should become True
+                    lpaCostsAppliedFor=False,  # Should become True
+                    changedDevelopmentDescription=False,  # Should become True
```
stef-solirius: Minor, but the comments above say "Should become True" for False inputs

odw/core/etl/transformation/curated/dart_api_curation_process.py
HarrisonBoyleThomas: Good spot, updated

odw/test/integration_test/etl/curated/test_dart_api_curation_process.py
HarrisonBoyleThomas: good spot, updated

odw/test/integration_test/etl/curated/test_dart_api_curation_process.py
HarrisonBoyleThomas: Updated, although the schemas are the same

odw/test/integration_test/etl/curated/test_dart_api_curation_process.py
HarrisonBoyleThomas: Let me double check, the final data was generated by running the original process in the notebook

odw/test/integration_test/etl/curated/test_dart_api_curation_process.py
HarrisonBoyleThomas: Good spot, updated

odw/test/unit_test/etl/curated/test_dart_api_curation_process.py
HarrisonBoyleThomas: Updated

odw/test/unit_test/etl/curated/test_dart_api_curation_process.py
HarrisonBoyleThomas: Updated

odw/test/integration_test/etl/curated/test_dart_api_curation_process.py
HarrisonBoyleThomas: Ok so i think it was going to pass as it was, but upon closer inspection it looked like i'd accidentally left the test data with `IsActive = None` so all of those fields were being accidentally nullified. I've updated the test data so now it flows properly

stef-solirius (approved):
LGTM, thanks Harrison.

PR #2497, opened 2026-05-05
feat: THEODW-3081 Restructure EntraID Notebooks in Standardised and Harmonised Layer

stef-solirius:
https://pins-ds.atlassian.net/browse/THEODW-3081

## Summary

- Added `entraid_standardisation_process.py` and updated `entraid_harmonisation_process.py` as structured ETL processes replacing notebook-based logic for EntraID data in the standardised and harmonised layers
- Added new Synapse pipelines: `pln_api_ingest` and `rel_34_0_0_post_deployment` (release 34.0.0 post-deployment steps)
- Updated `pln_orchestrator` and `manage_main_pipeline_config_entry` to integrate the new pipeline structure
- Added unit and integration tests for both EntraID standardisation and harmonisation processes

## Test plan

- [ ] Unit tests pass: `poetry run pytest tests/unit_test/ -m "not integration"`
- [ ] Integration tests pass: `poetry run pytest tests/integration_test/`
- [ ] `pln_api_ingest` pipeline runs end-to-end in Development environment
- [ ] Orchestrator correctly routes to new EntraID ingest pipeline
- [ ] Release 34.0.0 post-deployment pipeline executes without errors
- [ ] Verify no regressions in existing AIE and API ingest flows


odw/core/etl/transformation/harmonised/entraid_harmonisation_process.py
```diff
@@ -1,20 +1,220 @@
...
+            F.col("IngestionDate"),
+            F.lit(None).cast("string").alias("ValidTo"),
+            F.col("RowID"),
+            F.lit("Y").alias("IsActive"),
+        )
+
+        # Step 4: Close old versions of changed records
+        LoggingUtil().log_info("Closing old versions of changed records")
+        changed_ids = candidates.filter(F.col("HistoricIsActive") == "Y").select("id")
+        closing_date = F.to_timestamp(F.date_sub(F.current_date(), 1))
```
Fred83200: The notebook closes old active rows using the new record IngestionDate minus 1 day not the current_date minus 1

It does this in the notebook : 
to_timestamp(date_sub(IngestionDate, 1)) AS ClosingDate
if you use current_date here you'll produce a different ValidTo

odw/core/etl/transformation/harmonised/entraid_harmonisation_process.py
```diff
@@ -1,20 +1,220 @@
...
+        old_versions = (
+            hrm_active.join(changed_ids, "id")
+            .withColumn("IsActive", F.lit("N"))
+            .withColumn("ValidTo", closing_date.cast("string"))
```
Fred83200: Should it be cast to string ? The notebook creates ClosingDate as a timestamp

odw/core/etl/transformation/harmonised/entraid_harmonisation_process.py
```diff
@@ -1,20 +1,220 @@
...
+                F.col("std.userPrincipalName"),
+                F.col("std.IngestionDate"),
+                F.col("std.RowID"),
+                F.col("hrm.IsActive").alias("HistoricIsActive"),
+            )
+        )
+
+        # Step 3: New rows to insert with IsActive='Y'
+        LoggingUtil().log_info("Building new/updated rows")
+        output_cols = [
```
Fred83200, line 137: Here you need to double check the col order. 
in the notebook it does : 
EmployeeEntraId, employeeId, id, givenName.....

Here it does: 
EmployeeEntraId, id, employeeId, givenName.... 
Just saying this cause the write is overwriteSchema=true so if it is not in the same order it will change the table

odw/test/unit_test/etl/harmonised/test_entraid_harmonisation_process.py
```diff
@@ -0,0 +1,269 @@
...
+            [_hrm_row("user2", "emp2", "OldFirst", "Last2", "user2@test.com", is_active="Y", employee_entra_id=1)],
+            _HRM_SCHEMA,
+        )
+        data_to_write, _ = self._run_process(std_df, hrm_df)
+        df = data_to_write[EntraIdHarmonisationProcess.OUTPUT_TABLE]["data"]
+        assert df.count() == 2
+        by_active = {row["IsActive"]: row for row in df.collect()}
+        assert by_active["Y"]["givenName"] == "NewFirst"
+        assert by_active["N"]["givenName"] == "OldFirst"
+        assert by_active["N"]["ValidTo"] is not None
```
Fred83200: Can we assert that the old row ValidTo equals new IngestionDate minus one day? It's what the notebook was doing

Fred83200: Something like this prob : 
assert by_active["N"]["ValidTo"] == datetime(2024, 1, 31)

odw/test/integration_test/etl/harmonised/test_entraid_harmonisation_process.py
Fred83200, line 1: I think for both the integration test you could reduce the current integration suite to a couple of run() or
add a more end to end style integration test that uses the real load_data path instead of mocking it. 
Basically your int test should validate larger e2e legacy behaviour rather than individual transformation which your unit test cover

odw/core/etl/transformation/harmonised/entraid_harmonisation_process.py
```diff
@@ -1,20 +1,220 @@
...
+        """)
+
+        LoggingUtil().log_info(f"Loading current harmonised data from {self.HRM_TABLE}")
+        hrm_data = self.spark.sql(f"SELECT * FROM {self.HRM_TABLE}")
+
+        LoggingUtil().log_info(f"Loading source system fact from {self.SOURCE_SYSTEM_TABLE}")
+        source_system = self.spark.sql(f"""
+            SELECT SourceSystemID
+            FROM {self.SOURCE_SYSTEM_TABLE}
+            WHERE Description = 'SAP HR' AND IsActive = 'Y'
```
Fred83200, line 77: This to me is a business issue. 
It matches the notebook so good on that but I don't understand why the output ODTSourceSystem is 'EntraID' but SourceSystemID is looked up from Description = 'SAP HR'

workspace/pipeline/pln_aie_ingest.json
```diff
@@ -0,0 +1,528 @@
...
+					]
+				}
+			},
+			{
+				"name": "Append success",
+				"type": "AppendVariable",
+				"dependsOn": [
+					{
+						"activity": "For Each AIE Document",
+						"dependencyConditions": [
```
Fred83200, line 340: The append success only runs when the ForEach succeeds overall which mean if one file copy fails we only append the failure and not partial success. 
It's fine by me if that's what you want though just wondering

workspace/pipeline/pln_aie_ingest.json
```diff
@@ -0,0 +1,528 @@
...
+							"Succeeded"
+						]
+					}
+				],
+				"userProperties": [],
+				"typeProperties": {
+					"variableName": "aieStatus",
+					"value": {
+						"value": "@json(concat(\n  '{',\n    '\"entity\":\"AIE Documents\",',\n    '\"status\":\"Succeeded\",',\n    '\"recordsIngested\":0,',\n    '\"durationSeconds\":', string(div(sub(ticks(utcNow()), int(variables('loopStartTicks'))), 10000000)), ',',\n    '\"runId\":\"', pipeline().RunId, '\"',\n  '}'\n))",
```
Fred83200: recordsIngested is hardcoded to 0 is that intentional ? guessing you copy files and dont know the row count right at this stage ? 
Thinking if it's worth adding a 'filesCopied' field so the parent knows how many file were ingested maybe ?

workspace/pipeline/pln_api_ingest.json
```diff
@@ -0,0 +1,689 @@
...
+									"dependencyConditions": [
+										"Succeeded"
+									]
+								}
+							],
+							"userProperties": [],
+							"typeProperties": {
+								"variableName": "apiStatus",
+								"value": {
+									"value": "@first(json(activity('AIE_Sub').output.pipelineReturnValue.result).results)",
```
Fred83200, line 280: Should this append only the first AIE result ? from what I understand the pipeline returns results as an array but here you use first(....) so if you ever return multiple objects you only keep the first one. 
If it's expected to return a single summary object it's ok for me tho just thinking if not you might lose some details

workspace/pipeline/pln_api_ingest.json
```diff
@@ -0,0 +1,689 @@
...
+									"dependencyConditions": [
+										"Succeeded"
+									]
+								}
+							],
+							"userProperties": [],
+							"typeProperties": {
+								"variableName": "apiStatus",
+								"value": {
+									"value": "@json(concat(\n  '{',\n    '\"entity\":\"Listed Buildings\",',\n    '\"status\":\"Succeeded\",',\n    '\"recordsIngested\":0,',\n    '\"durationSeconds\":', string(div(sub(ticks(utcNow()), int(variables('loopStartTicks'))), 10000000)), ',',\n    '\"runId\":\"', pipeline().RunId, '\"',\n  '}'\n))",
```
Fred83200: Same Q here as earlier : recordsIngested set to 0 for Listed Building. Should we capture rows copied from LB_Copy_Buildings or LB_Copy_Outlines ?

workspace/pipeline/pln_orchestrator.json
```diff
@@ -1482,6 +1662,10 @@
 				"type": "bool",
 				"defaultValue": true
 			},
+			"runIngestAPI": {
```
Fred83200, line 1697: This param is duplicated

workspace/pipeline/pln_vault_and_env_init.json
```diff
@@ -691,6 +697,98 @@
...
+									]
+								}
+							],
+							"policy": {
+								"secureOutput": false,
+								"secureInput": false
+							},
+							"userProperties": [],
+							"typeProperties": {
+								"variableName": "service_bus_feed",
```
Fred83200: You Set api feed is stored in service_bus_feed

Fred83200: You might need to add another var to store it into like api_feed

workspace/pipeline/pln_vault_and_env_init.json
```diff
@@ -691,6 +697,98 @@
...
+				],
+				"userProperties": [],
+				"typeProperties": {
+					"expression": {
+						"value": "@pipeline().parameters.includeAPI\n",
+						"type": "Expression"
+					},
+					"ifFalseActivities": [
+						{
+							"name": "Init Service Bus_copy1",
```
Fred83200: I'd maybe change this name to match API

workspace/pipeline/pln_vault_and_env_init.json
```diff
@@ -691,6 +697,98 @@
 						}
 					}
 				}
+			},
+			{
+				"name": "If_API",
+				"description": "Check if includeServiceBus is True or False",
```
Fred83200: I'd change the description to match API

workspace/pipeline/rel_34_0_0_post_deployment.json
Fred83200, line 1: after Delete_Horizon_Entries I think it might be safe to chain all the config updates sequentially instead of running them in parallel. 
Even if it looks like each bit is safe to run in parallel they all run some update/merge against the same table at the same time which might create a race condition which will create a failure..

odw/core/etl/transformation/harmonised/entraid_harmonisation_process.py
stef-solirius: Thanks. Fixed output_cols and new_rows column order: employeeId now comes before id (matching the notebook)

odw/core/etl/transformation/harmonised/entraid_harmonisation_process.py
stef-solirius: True, addressed with work above.

odw/core/etl/transformation/harmonised/entraid_harmonisation_process.py
stef-solirius: Addressed. ClosingDate now uses to_timestamp(date_sub(new_IngestionDate, 1)) from the incoming record, not current_date - 1

odw/test/unit_test/etl/harmonised/test_entraid_harmonisation_process.py
stef-solirius: Thanks. Added exact ValidTo assertion: "2024-01-31 00:00:00" (IngestionDate 2024-02-01 minus one day)

odw/test/integration_test/etl/harmonised/test_entraid_harmonisation_process.py
stef-solirius: I did a full rewrite: now it uses real load_data via temp views + patched table constants, mocks only write_data - 4 E2E tests covering new record, SCD2 close, unchanged, and historical

workspace/pipeline/pln_orchestrator.json
stef-solirius: Removed thanks

workspace/pipeline/pln_vault_and_env_init.json
stef-solirius: Actioned: 
- Renamed activity Init Service Bus_copy1 → Init API feed                                                                                                            
  - Fixed description of If_API: "Check if includeAPI is True or False"                                                                                                
  - Both API-feed activities now write to api_feed variable (not service_bus_feed)                                                                                     
  - Added api_feed variable declaration                                                                                                                                
  - Added apiEntities to the Return outputs pipeline return value

workspace/pipeline/pln_vault_and_env_init.json
stef-solirius: Actioned

workspace/pipeline/pln_vault_and_env_init.json
stef-solirius: Actioned

workspace/pipeline/rel_34_0_0_post_deployment.json
stef-solirius: True, I chained post-deploy activities sequentially: Delete_Horizon_Entries → Rename_Listed_Buildings_To_API → Rename_HR_To_API → Rename_Entraid_Entity_Name →             
  Add_AIE_Document_Data_Entry

workspace/pipeline/pln_api_ingest.json
stef-solirius: Actioned. LB_Append_Success now reports actual recordsIngested as the sum of LB_Copy_Buildings.rowsCopied + LB_Copy_Outlines.rowsCopied

workspace/pipeline/pln_api_ingest.json
stef-solirius: Fixed. Added Append forEach failure activity triggered when For Each AIE Document fails

workspace/pipeline/pln_aie_ingest.json
stef-solirius: Added filesCopied field to the success append using metadata count

workspace/pipeline/pln_aie_ingest.json
stef-solirius: Fixed. Added Append forEach failure activity triggered when For Each AIE Document fails

odw/core/etl/transformation/harmonised/entraid_harmonisation_process.py
stef-solirius: We'll keep EntraID now

odw/test/integration_test/etl/harmonised/test_entraid_harmonisation_process.py
```diff
@@ -0,0 +1,169 @@
...
+        T.StructField("id", T.StringType(), True),
+        T.StructField("employeeId", T.StringType(), True),
+        T.StructField("givenName", T.StringType(), True),
+        T.StructField("surname", T.StringType(), True),
+        T.StructField("userPrincipalName", T.StringType(), True),
+        T.StructField("expected_from", T.StringType(), True),
+    ]
+)
+
+_HRM_SCHEMA = T.StructType(
```
Fred83200: You changed the output_cols which is good but now the test is not matching the real output col order. So you might need to adjust it there too and other test

odw/core/etl/transformation/harmonised/entraid_harmonisation_process.py
Fred83200: Canm we add a small comment explaining why EntraID uses the SAP HR source system lookup while ODTSourceSystem is set to EntraID It looks confusing without the notebook context

workspace/pipeline/pln_aie_ingest.json
```diff
@@ -0,0 +1,548 @@
...
+				"typeProperties": {
+					"variableName": "aieStatus",
+					"value": {
+						"value": "@json(concat(\n  '{',\n    '\"entity\":\"AIE Documents\",',\n    '\"status\":\"Succeeded\",',\n    '\"recordsIngested\":0,',\n    '\"filesCopied\":', string(length(activity('Get Metadata Of CSVs').output.childItems)), ',',\n    '\"durationSeconds\":', string(div(sub(ticks(utcNow()), int(variables('loopStartTicks'))), 10000000)), ',',\n    '\"runId\":\"', pipeline().RunId, '\"',\n  '}'\n))",
+						"type": "Expression"
+					}
+				}
+			},
+			{
+				"name": "Append forEach failure",
```
Fred83200: I think you have a duplciate now for the failure handling. 
Append forEach failure and Append failure foreach both trigger when For Each AIE Document fails and both append a failed AIE result. 
I think you need to remove one of them to avoid duplicated failure entry.

workspace/pipeline/pln_api_ingest.json
Fred83200: Problem is that this is the parent pipeline with this expression using first() the parent pipeline only appends the first object from AIE child pipeline results array
If its is guaranteed to return one summary object only then it's fine. If not and can return multiple objects for example a success + a failure you will lose anything after the first result

odw/core/etl/transformation/standardised/entraid_standardisation_process.py
```diff
@@ -0,0 +1,143 @@
...
+        )
+        if not folders:
+            raise FileNotFoundError(f"No date folders found under {raw_path}")
+        return folders[0]
+
+    # ------------------------------------------------------------------
+    # load_data – resolves the folder and reads the raw JSON file
+    # ------------------------------------------------------------------
+
+    def load_data(self, **kwargs) -> Dict[str, DataFrame]:
```
Fred83200, line 59: One difference here from the notebook. The notebook reads the latest raw EntraID file then creates raw_entraids and then joins entraid_new_candidates back to raw_entraids before building entraid_new. 
It looks like here we only use the std table, current hrm table and source system table during the process. 
If STD is now the source of truth it's fine by me but is the difference intentional ? Just double checking as the logic in the notebook uses the raw snapshot to remove duplicates / validate current records before creating entraid_new

odw/core/etl/transformation/harmonised/entraid_harmonisation_process.py
stef-solirius: Added

workspace/pipeline/pln_api_ingest.json
stef-solirius: Both changes are applied: 
  - I removed Append forEach failure (the duplicate activity that caused a double-append on ForEach failure). Append failure foreach + Mark Failure foreach remain as the canonical failure path.                                                                                                         
  -  AIE_Append_Status changed from AppendVariable + first() to SetVariable + union(variables('apiStatus'), ...), so all items from the child's results array are merged in regardless of how many there are.

odw/test/integration_test/etl/harmonised/test_entraid_harmonisation_process.py
stef-solirius: Fixed, _HRM_SCHEMA and _hrm_row() now have employeeId before id in both the unit and integration test, matching the output_cols order in the process

workspace/pipeline/pln_aie_ingest.json
stef-solirius: Append forEach is gone now, only Append failure meta (triggered by Get Metadata Of CSVs failing) and Append failure foreach (triggered by the ForEach failing) remain, which are two distinct failure scenarios.

odw/core/etl/transformation/standardised/entraid_standardisation_process.py
stef-solirius: You're right, good catch. I've fixed that

stef-solirius: On second thoughts... The notebook needed this because the STD table historically accumulated data across multiple runs (not necessarily overwritten each time), so the raw snapshot was used to constrain entraid_new_candidates to only records from this specific run.... I cannot think anymore of a scenario where we will harmonisation on stale standardisation. I might be wrong though

Fred83200: Make sense - As long as STD now represent the latest valid snapshot for HRM then I'm fine with dropping that step

Fred83200 (commented):
The fixes look good now. 

Before approving , the tests should be more align with appeal_has "new" pattern. 

What I mean is : 
- use schema helper functions like _hrm_schema(), _std_schema(), _raw_entraid_schema() instead of module-level schema constants
- add a _source_data(...) helper so the setup is consistent and cleaner
keep unit tests focused on process()
- keep integration tests focused on run() by mocking load_data and write_data, rather than creating temp views and exercising the SQL inside load_data
- use assert_dataframes_equal() where we compare actual vs expected output, instead of lots of individual row asserts
- add an explicit output column order assertion for harmonised, e.g. df.columns == [field.name for field in _hrm_schema()], because this already caught us once
- avoid patching LoggingUtil unless it is actually required, as the base test classes already handle the common test setup (LoggingUtil come from SparkTestCase & ETLTestCase inherits from it)

odw/test/integration_test/etl/standardised/test_entraid_standardisation_process.py
```diff
@@ -0,0 +1,113 @@
...
+
+import mock
+import odw.test.util.mock.import_mock_notebook_utils  # noqa: F401
+import pyspark.sql.types as T
+
+from odw.core.etl.transformation.standardised.entraid_standardisation_process import EntraIdStandardisationProcess
+from odw.test.integration_test.etl.etl_test_case import ETLTestCase
+from odw.test.util.session_util import PytestSparkSessionUtil
+
+_VALUE_SCHEMA = T.StructType(
```
Fred83200: I'd change that into helper functions like _value_schema()

odw/test/integration_test/etl/standardised/test_entraid_standardisation_process.py
```diff
@@ -0,0 +1,113 @@
...
+    [
+        T.StructField("id", T.StringType(), True),
+        T.StructField("employeeId", T.StringType(), True),
+        T.StructField("givenName", T.StringType(), True),
+        T.StructField("surname", T.StringType(), True),
+        T.StructField("userPrincipalName", T.StringType(), True),
+    ]
+)
+
+_RAW_ENTRAID_SCHEMA = T.StructType(
```
Fred83200: Same for this one : _raw_entraid_schema()

odw/test/integration_test/etl/standardised/test_entraid_standardisation_process.py
```diff
@@ -0,0 +1,113 @@
...
+def _raw_df(spark, records):
+    return spark.createDataFrame([([dict(r) for r in records],)], _RAW_ENTRAID_SCHEMA)
+
+
+class TestEntraIdStandardisationProcessIntegration(ETLTestCase):
+    def _run(self, raw_df, folder_name="2024-01-15"):
+        spark = PytestSparkSessionUtil().get_spark_session()
+        source_data = {"raw_entraid": raw_df}
+
+        with (
```
Fred83200: You can also remove all the patching for LoggingUtil throughout the test

odw/test/unit_test/etl/standardised/test_entraid_standardisation_process.py
```diff
@@ -0,0 +1,159 @@
+from datetime import datetime
+
+import mock
+import pyspark.sql.types as T
+
+from odw.core.etl.transformation.standardised.entraid_standardisation_process import EntraIdStandardisationProcess
+from odw.test.util.session_util import PytestSparkSessionUtil
+from odw.test.util.test_case import SparkTestCase
+
+_VALUE_SCHEMA = T.StructType(
```
Fred83200: Same here - This can change into a helper function

odw/test/unit_test/etl/standardised/test_entraid_standardisation_process.py
```diff
@@ -0,0 +1,159 @@
...
+    [
+        T.StructField("id", T.StringType(), True),
+        T.StructField("employeeId", T.StringType(), True),
+        T.StructField("givenName", T.StringType(), True),
+        T.StructField("surname", T.StringType(), True),
+        T.StructField("userPrincipalName", T.StringType(), True),
+    ]
+)
+
+_RAW_ENTRAID_SCHEMA = T.StructType(
```
Fred83200: Same for this one

odw/test/unit_test/etl/standardised/test_entraid_standardisation_process.py
```diff
@@ -0,0 +1,159 @@
...
+
+class TestEntraIdStandardisationProcess(SparkTestCase):
+    def _run_process(self, raw_df, folder_name="2024-01-15"):
+        spark = PytestSparkSessionUtil().get_spark_session()
+        with (
+            mock.patch(
+                "odw.core.etl.transformation.standardised.entraid_standardisation_process.Util.get_storage_account",
+                return_value="test_storage",
+            ),
+            mock.patch("odw.core.etl.transformation.standardised.entraid_standardisation_process.LoggingUtil"),
```
Fred83200: No need to patch LoggingUtil  - You can remove it throughout the test

odw/test/integration_test/etl/standardised/test_entraid_standardisation_process.py
```diff
@@ -0,0 +1,113 @@
...
+    ]
+)
+
+
+def _raw_df(spark, records):
+    return spark.createDataFrame([([dict(r) for r in records],)], _RAW_ENTRAID_SCHEMA)
+
+
+class TestEntraIdStandardisationProcessIntegration(ETLTestCase):
+    def _run(self, raw_df, folder_name="2024-01-15"):
```
Fred83200, line 41: Maybe I'd add a small _source_data(raw_df) helper like the HRM tests use just to keep it consistent

odw/test/unit_test/etl/standardised/test_entraid_standardisation_process.py
```diff
@@ -0,0 +1,159 @@
...
+        spark = PytestSparkSessionUtil().get_spark_session()
+        data_to_write, result = self._run_process(_empty_raw_df(spark))
+        assert data_to_write[EntraIdStandardisationProcess.OUTPUT_TABLE]["data"].count() == 0
+        assert result.metadata.insert_count == 0
+
+    # ------------------------------------------------------------------
+    # process – schema
+    # ------------------------------------------------------------------
+
+    def test__process__output_contains_expected_columns(self):
```
Fred83200: might be good to add an explicit output column order assert somewhere like : 
assert df.columns == ["id", "employeeId", "givenName", "surname", "userPrincipalName", "expected_from"]

mainly cause you had hit a col order issue once in HRM

odw/test/integration_test/etl/standardised/test_entraid_standardisation_process.py
stef-solirius: Actioned

odw/test/integration_test/etl/standardised/test_entraid_standardisation_process.py
stef-solirius: Actioned

odw/test/integration_test/etl/standardised/test_entraid_standardisation_process.py
stef-solirius: Actioned

odw/test/unit_test/etl/standardised/test_entraid_standardisation_process.py
stef-solirius: Actioned

odw/test/unit_test/etl/standardised/test_entraid_standardisation_process.py
stef-solirius: Actioned

odw/test/unit_test/etl/standardised/test_entraid_standardisation_process.py
stef-solirius: Actioned

odw/test/integration_test/etl/standardised/test_entraid_standardisation_process.py
stef-solirius: Actioned

odw/test/unit_test/etl/standardised/test_entraid_standardisation_process.py
stef-solirius: Done and replaced :)

Fred83200 (dismissed):

Fred83200 (approved):

PR #2505, opened 2026-05-11
Feat/theodw 3103 refine integration tests2

HarrisonBoyleThomas:
Ticket: https://pins-ds.atlassian.net/browse/THEODW-3103

- Update appeal has integration tests to test that input data correctly flows through the ETL process
- Migrate old some integration tests to the unit test folder (since there were more like unit tests than integration tests, and it is good to keep these) - Not all integration tests were migrated because there was overlapping logic with the existing unit tests


Fred83200 (approved):
LGTM

PR #2517, opened 2026-05-18
Feat/theodw 2962

Fred83200:
[THEODW-2962](https://pins-ds.atlassian.net/browse/THEODW-2962)

Added unit and integration tests for appeal_event HRM/CUR  covering legacy notebook behaviour including active row filtering, column projection/renaming, overwrite behaviour, duplicate handling, empty inputs, and curated output validation.

[THEODW-2962]: https://pins-ds.atlassian.net/browse/THEODW-2962?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


odw/test/integration_test/etl/curated/test_appeal_event_curated_process.py
```diff
@@ -0,0 +1,313 @@
...
+from odw.core.etl.transformation.curated.appeal_event_curated_process import AppealEventCuratedProcess
+from odw.test.integration_test.etl.etl_test_case import ETLTestCase
+from odw.test.util.assertion import assert_dataframes_equal, assert_etl_result_successful
+from odw.test.util.session_util import PytestSparkSessionUtil
+
+
+pytestmark = pytest.mark.xfail(reason="Curated logic not implemented yet")
+
+
+CURATED_COLUMNS = [
```
stef-solirius, line 17: I might be wrong, but shouldn't this also be an helper function, rather than a constant?
e.g. def _cur_schema():

Fred83200: Kept it as a constant since its a fixed static column projection used for assertions rather than generated dynamically

odw/test/unit_test/etl/curated/test_appeal_event_curated_process.py
```diff
@@ -0,0 +1,288 @@
...
+from odw.core.etl.transformation.curated.appeal_event_curated_process import AppealEventCuratedProcess
+from odw.test.util.assertion import assert_dataframes_equal
+from odw.test.util.session_util import PytestSparkSessionUtil
+from odw.test.util.test_case import SparkTestCase
+
+
+pytestmark = pytest.mark.xfail(reason="Curated logic not implemented yet")
+
+
+CURATED_COLUMNS = [
```
stef-solirius, line 16: Same here, should we we make this an helper function? def def _cur_schema():

odw/test/unit_test/etl/harmonised/test_appeal_event_harmonisation_process.py
```diff
@@ -0,0 +1,376 @@
...
+from odw.core.etl.transformation.harmonised.appeal_event_harmonisation_process import AppealEventHarmonisationProcess
+from odw.test.util.assertion import assert_dataframes_equal
+from odw.test.util.session_util import PytestSparkSessionUtil
+from odw.test.util.test_case import SparkTestCase
+
+
+pytestmark = pytest.mark.xfail(reason="Harmonised logic not implemented yet")
+
+
+HARMONISED_COLUMNS = [
```
stef-solirius, line 16: Same here def _harmonised_schema():

odw/test/unit_test/etl/curated/test_appeal_event_curated_process.py
```diff
@@ -0,0 +1,288 @@
...
+
+def _harmonised_schema():
+    return T.StructType(
+        [
+            T.StructField("eventId", T.StringType(), True),
+            T.StructField("caseReference", T.StringType(), True),
+            T.StructField("eventType", T.StringType(), True),
+            T.StructField("eventName", T.StringType(), True),
+            T.StructField("eventStatus", T.StringType(), True),
+            T.StructField("IsUrgent", T.BooleanType(), True),
```
stef-solirius: This uses IsUrgent (capital I), while the harmonised integration test's _existing_harmonised_df schema uses isUrgent (lowercase). I think he might fail

odw/test/unit_test/etl/harmonised/test_appeal_event_harmonisation_process.py
```diff
@@ -0,0 +1,376 @@
...
+import mock
+import pytest
+import pyspark.sql.types as T
+from odw.core.etl.transformation.harmonised.appeal_event_harmonisation_process import AppealEventHarmonisationProcess
+from odw.test.util.assertion import assert_dataframes_equal
+from odw.test.util.session_util import PytestSparkSessionUtil
+from odw.test.util.test_case import SparkTestCase
+
+
+pytestmark = pytest.mark.xfail(reason="Harmonised logic not implemented yet")
```
stef-solirius: Consider adding raises=NotImplementedError to xfail so that if the test fails for a different reason (assertion error rather than stub) it is reported as an XPASS/unexpected failure rather than silently accepted.

odw/test/integration_test/etl/harmonised/test_appeal_event_harmonisation_process.py
```diff
@@ -0,0 +1,423 @@
...
+    "ValidTo",
+    "RowID",
+    "IsActive",
+]
+
+
+def _service_bus_schema():
+    return T.StructType(
+        [
+            T.StructField("AppealsEventId", T.LongType(), True),
```
stef-solirius, line 47: Here is a LongType in _service_bus_schema but IntegerType in _existing_harmonised_df. If the union/merge writes both into the same output schema, a type mismatch will surface at runtime. One of the two needs correcting; LongType is safer for auto-incremented IDs imo.

odw/test/integration_test/etl/harmonised/test_appeal_event_harmonisation_process.py
```diff
@@ -0,0 +1,423 @@
...
+
+class TestAppealEventHarmonisationProcess(ETLTestCase):
+    def _run_process(self, spark, test_case: str):
+        service_bus_table = f"{test_case}_sb_appeal_event"
+        horizon_table = f"{test_case}_horizon_appeals_event"
+        output_table = f"{test_case}_appeal_event"
+
+        self.write_existing_table(
+            spark,
+            _service_bus_df(spark),
```
stef-solirius: Not sure why we are hardcoding the sb data here (in curated you were passing the source_df as a parameter). This makes it impossible to write an integration-level "empty sources" test without duplicating the entire setup block. Accept service_bus_df and horizon_df as parameters, mirroring the curated test pattern, see here https://github.com/Planning-Inspectorate/odw-synapse-workspace/blob/main/odw/test/integration_test/etl/harmonised/test_appeal_document_harmonisation_process.py

odw/test/integration_test/etl/curated/test_appeal_event_curated_process.py
```diff
@@ -0,0 +1,313 @@
...
+        assert "extraColumn" not in actual_df.columns
+        assert "notificationOfSitevisit" not in actual_df.columns
+        assert "notificationOfSiteVisit" in actual_df.columns
+
+        assert {
+            "insert_count": result.metadata.insert_count,
+            "update_count": result.metadata.update_count,
+            "delete_count": result.metadata.delete_count,
+        } == {
+            "insert_count": 1,
```
stef-solirius: _assert_curation hardcodes insert_count: 1, so the empty-input test can't call it and must duplicate the count assertions. Either make expected_counts a parameter, or extract the column-shape assertions into a separate helper that both tests can call.

odw/test/unit_test/etl/curated/test_appeal_event_curated_process.py
Fred83200: Good catch - Fixed the typo

odw/test/unit_test/etl/harmonised/test_appeal_event_harmonisation_process.py
Fred83200: Added it for all tests

odw/test/integration_test/etl/harmonised/test_appeal_event_harmonisation_process.py
Fred83200: Yh changed it to LongType

odw/test/integration_test/etl/harmonised/test_appeal_event_harmonisation_process.py
Fred83200: Changed

odw/test/integration_test/etl/curated/test_appeal_event_curated_process.py
Fred83200: Done

stef-solirius (approved):
LGTM

PR #2519, opened 2026-05-18
Feat/3003 incremental processing pln service bus 1805

KranthiRayipudi:
[https://pins-ds.atlassian.net/browse/THEODW-3003?atlOrigin=eyJpIjoiMTliOTJlOGExMzFiNGU3ZmJkZjdkMWFlNmRmMzE1YTAiLCJwIjoiaiJ9](url)

•	Added new notebook py_sb_get_last_load_date to fetch last processed timestamp and count new Service Bus messages 
•	Added new notebook nb_sb_watermark to create and maintain watermark table using Delta MERGE 
•	Introduced watermark table odw_meta_db.pln_service_bus_watermarks to store last_ingest_time per entity 
•	Enhanced pln_service_bus pipeline with incremental processing logic 
•	Added step to calculate number of new messages before processing 
•	Introduced If condition to execute pipeline only when new data is available 
•	Skips processing and logs warning when no new messages are found 
•	Integrated dynamic orchestration using orchestration_ServiceBus.json 
•	Added support for entity-specific harmonised notebook execution 
•	Added special handling for nsip-project entity 
•	Integrated watermark update at the end of pipeline execution 
•	Improved logging using pln_log_to_appins for better monitoring 
•	Reduced unnecessary compute by avoiding processing when no new messages 
•	Ensured idempotent and incremental ingestion for Service Bus entities

**PR Template**

 1. JIRA Ticket Reference  :
         
    [ Enter JIRA ticket number and Title here]

 2.  Summary of the work : 
         
    [ Enter Summary here]
 
 3.  New Source-to-Raw Datasets
	
	 - [ ] New source data has been added
  	    - A trigger has been attached at the appropriate frequency

 4.  New Tables in Standardised Layer

   	 - [ ] New standardised tables have been created
		 -  orchestration.json is updated and tested in Dev, and PR is open or merged to main
		 -  Schema exists in odw-config/standardised-table-definitions or is about to be PRd
 
 5.   New Tables in Harmonised or Curated Layers

      - [ ] New harmonised or curated tables have been created
 		- Script is configured in the pipeline pln_post_deployments
        - Schema exists in odw-config/harmonised-table-definitions or curated-table-definitions or is about to be PRd
 
 6.  Schema or Column Changes
    (Only new columns or columns with changed data types are in scope)

     - [ ] Changes to table structure or columns
		  - py_change_table is set to run in pln_post_deployments
	      - A script has been created to backfill or populate new column(s) in Test and Prod
			 - [ ] Avoid dropping and recreating tables unless strictly necessary

 7. Script Execution in Build
	 - [ ] Scripts have run in isolation in Build
		-  Script has been added to pln_post_deployments
	    -  Script is now part of a scheduled pipeline with correct triggers
	 - [ ] No scripts have run or no action required in Test/Prod

8. Table Creation and Schema Validation
   
  	 - [ ] All required tables have been created
  	 - [ ] Schema has been validated against the requirements

9.  Deployment and Schema Change Documentation
	
  	 - [ ] Deployment steps and rollback procedures are documented
  	 - [ ] Schema change handling is outlined and tested

10. Archiving Process Review

  	 - [ ] Automatic archiving logic has been reviewed
  	 - [ ] Archiving schedules and retention policies are validated


Fred83200 (commented):
This does not currently look aligned with the agreed watermark-based design. The implementation still appears to use file/path timestamp comparison through py_sb_get_last_load_date, which is the approach we were trying to avoid. The agreed design was to use message_enqueued_time_utc from RAW as the incremental reference, with the watermark table storing the last successfully processed enqueue time per entity.

I also do not think we should introduce a new pln_master_refactor pipeline as part of this ticket. The scope should be to update the existing pln_service_bus pipeline and keep the change focused.

I would suggest simplifying the implementation:

Keep one watermark notebook for creating/upserting the watermark table.
Remove py_sb_get_last_load_date.
Do not derive new data from filenames or folder timestamps.
Update pln_service_bus directly.
Ensure RAW to STD reads only records where message_enqueued_time_utc is greater than the stored watermark.
Update the watermark only after the successful STD > HRM processing chain.

At the moment this PR mixes the rejected file based approach with the watermark approach, which makes the design harder to maintain and less aligned with the agreed ODW direction.

KranthiRayipudi:
As you mentioned, I just want to confirm if my understanding is correct or not.
We need to remove py_sb_get_last_load_date and instead use message_enqueued_time_utc from the RAW layer.
 Option 1: Notebook Approach
For this, I believe we need to:
•	Read all JSON files recursively from the RAW folder for that entity
•	Extract message_enqueued_time_utc from the files
•	Compare it with the watermark table value (last_ingest_time)
If records are found where:
message_enqueued_time_utc > last_ingest_time
Then:
•	Proceed to execute the RAW → STD notebook
Otherwise:
•	Skip execution
Also, based on this approach, I think we need to create a notebook to handle this filtering and comparison logic.

 Option 2: Pipeline (Data Flow) Approach
Instead of using a notebook, we can implement this fully in the pipeline:
1.	Lookup → Read orchestration_service_bus.json for the entity(get raw path)
2.	Lookup → Get last_ingest_time from the watermark table
3.	Set Variable → RAW path
4.	Data Flow Activity:
o	Read RAW JSON files recursively
o	Filter: 
o	message_enqueued_time_utc > last_ingest_time
o	Aggregate → count records
o	Output → record_count
5.	If Condition:
o	If record_count > 0 → execute RAW → STD processing
o	Else → skip execution
•	**One more clarification:** if we pick up raw_entity_path from orchestration_service_bus.json, are we expecting values like the following?

"raw_entity_path": "abfss://odw-raw@pinsstodwdevuks9h80mb.dfs.core.windows.net/ServiceBus/service-user/"

Fred83200:
> As you mentioned, I just want to confirm if my understanding is correct or not. We need to remove py_sb_get_last_load_date and instead use message_enqueued_time_utc from the RAW layer. Option 1: Notebook Approach For this, I believe we need to: • Read all JSON files recursively from the RAW folder for that entity • Extract message_enqueued_time_utc from the files • Compare it with the watermark table value (last_ingest_time) If records are found where: message_enqueued_time_utc > last_ingest_time Then: • Proceed to execute the RAW → STD notebook Otherwise: • Skip execution Also, based on this approach, I think we need to create a notebook to handle this filtering and comparison logic.
> 
> Option 2: Pipeline (Data Flow) Approach Instead of using a notebook, we can implement this fully in the pipeline:
> 
> 1. Lookup → Read orchestration_service_bus.json for the entity(get raw path)
> 2. Lookup → Get last_ingest_time from the watermark table
> 3. Set Variable → RAW path
> 4. Data Flow Activity:
>    o	Read RAW JSON files recursively
>    o	Filter:
>    o	message_enqueued_time_utc > last_ingest_time
>    o	Aggregate → count records
>    o	Output → record_count
> 5. If Condition:
>    o	If record_count > 0 → execute RAW → STD processing
>    o	Else → skip execution
>    •	**One more clarification:** if we pick up raw_entity_path from orchestration_service_bus.json, are we expecting values like the following?
> 
> "raw_entity_path": "abfss://[odw-raw@pinsstodwdevuks9h80mb.dfs.core.windows.net](mailto:odw-raw@pinsstodwdevuks9h80mb.dfs.core.windows.net)/ServiceBus/service-user/"

Hi Kranthi,

You are much closer to the intended design now, however I do not think we should introduce either:

- A separate pre-scan notebook purely to detect new data
- A Data Flow activity purely to count records before processing

Both approaches would result in reading/scanning RAW separately before the actual RAW > STD processing begins, which introduces unnecessary complexity and additional compute overhead.

The cleaner/scalable approach would be:

- Pipeline retrieves the entity watermark (`last_ingest_time`)
- Pipeline passes the watermark value into the existing RAW > STD notebook
- RAW > STD notebook reads RAW recursively for the entity
- RAW > STD notebook filters records using:
  `message_enqueued_time_utc > last_ingest_time`
- If the filtered dataframe is empty:

  - exit gracefully / skip downstream processing
- Otherwise:

  - continue RAW > STD
  - continue STD > HRM
  - update watermark only after successful completion

This avoids:

- Recursive RAW scans purely for detection
- Double reading RAW
- Additional Spark/Data Flow overhead
- File/path based orchestration logic

The incremental filtering should be part of the actual ingestion logic itself rather than a separate detection phase.

Regarding the orchestration config, I would not recommend storing full ABFSS paths such as:

`abfss://odw-raw@storage...`

That tightly couples configuration to environment specific storage accounts.

Instead, the config should contain only the logical/raw relative entity path, for example:

```
"raw_entity_path": "ServiceBus/service-user/"
```

and the full storage path should be constructed dynamically using the existing storage account utilities already used throughout ODW.

That keeps the implementation cleaner, portable across environments, and easier to maintain long term.

PR #2533, opened 2026-05-28
Feat/theodw 3112 dependency management

HarrisonBoyleThomas:
Ticket: https://pins-ds.atlassian.net/browse/THEODW-3112

- Add DependencyResolver which analyses the ETL config defined in `odw-config` to generate an execution order for ETLProcesses
- Rename `py_etl_orchestrator` to `py_etl_executor` (since this more accurately represents what it does). Update references
- Add `py_etl_orchestrator` which uses the dependency resolver to generate an execution order
- Add a generic orchestrator pipeline pair called `pln_etl_orchestrator` and `pln_etl_executor` which uses the dependency resolution logic to execute ETLProcess in parallel if possible and sequential when necessary
- Fix invalid import in `py_etl_executor`

Test evidence in the ticket


HarrisonBoyleThomas:
Closing because there is a bit more work to be done. I will reraise once ready

PR #2537, opened 2026-06-02
Feat/theodw 3112 dependency management

HarrisonBoyleThomas:
Ticket: https://pins-ds.atlassian.net/browse/THEODW-3112

- Add DependencyResolver which analyses the ETL config defined in `odw-config` to generate an execution order for ETLProcesses
- Add MetadataManager which stores an execution history for use in the new master pipeline
- Rename `py_etl_orchestrator` to `py_etl_executor` (since this more accurately represents what it does). Update references
- Add `py_etl_orchestrator` which uses the dependency resolver to generate an execution order
- Add a generic orchestrator pipeline pair called `pln_etl_orchestrator` and `pln_etl_executor` which uses the dependency resolution logic to execute ETLProcess in parallel if possible and sequential when necessary
- Fix invalid import in `py_etl_executor`

Test evidence in the ticket


workspace/notebook/py_etl_orchestrator.json
```diff
@@ -156,15 +180,16 @@
...
 			},
 			{
 				"cell_type": "code",
 				"source": [
-					"entity_stage_name = input_arguments.get(\"entity_stage_name\")\n",
-					"etl_process_class = ETLProcessFactory.get(entity_stage_name)\n",
-					"etl_process_instance = etl_process_class(spark, debug)"
+					"execution_history = []\n",
+					"if run_id:\n",
+					"    # Todo need to query the execution history DB. This should be filled in as part of the work to add the execution history table\n",
```
Fred83200: This atm wont work which is fine. 
if run_id is provided should this query odw_execution_history and pass actual execution details into DependencyResolver?

odw/core/orchestration/orchestration_config.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class OrchestrationConfig(BaseModel):
+    class EntityStage(BaseModel):
+        kwargs: Dict[str, Any]
+        depends_on: List[str]
+
+        @validator("kwargs")
+        def validate_kwargs(cls, v):
+            if "entity_stage_name" not in v:
```
Fred83200: What's the naming here ? 
Just to clarify - entity_stage_name is the ETL process name used by the ETLProcessFactory and the orchestration_entity_stage_name is generated later as the dependency key right ? 
Might be worth adding a comment or something to explain it

odw/core/orchestration/dependency_resolver.py
```diff
@@ -0,0 +1,102 @@
...
+        entity_stages = {f"{entity_name}.{stage_name}": stage for entity_name, entity in entities.items() for stage_name, stage in entity.items()}
+        visited = set()
+        unvisited = [x for x in entity_stages_to_keep]
+        while unvisited:
+            current = unvisited.pop()
+            unvisited.extend(entity_stages[current].get("depends_on", []))
+            visited.add(current)
+        cleaned_entities = dict()
+        for entity_stage_name in visited:
+            entity_stage_name_split = entity_stage_name.split(".")
```
Fred83200: Could this use rsplit(".", 1) instead ? it will avoid breaking if an entity or stage name even has a dot. Shouldn't happen though but we never know...

workspace/notebook/py_etl_executor.json
```diff
@@ -0,0 +1,247 @@
...
+					"# Validate notebook parameters"
+				]
+			},
+			{
+				"cell_type": "code",
+				"source": [
+					"required_keys = [\"entity_stage_name\", \"orchestration_run_id\", \"orchestration_entity_name\", \"orchestration_stage_name\"]\n",
+					"for key in required_keys:\n",
+					"    if key not in input_arguments:\n",
+					"        raise ValueError(f\"py_etl_ochestrator expects a '{key}' property to be supplied, but this was missing\")"
```
Fred83200: shouldn't it be py_etl_executor ? instead of orchestrator ?

workspace/notebook/py_etl_orchestrator.json
HarrisonBoyleThomas: Ah snap i forgot to commit this - let me add it in

odw/core/orchestration/orchestration_config.py
HarrisonBoyleThomas: Yeah that's right, but yeah agree. I think we should rename `entity_stage_name` to `etl_process_name`

odw/core/orchestration/dependency_resolver.py
HarrisonBoyleThomas: I'd prefer to ensure this isn't possible since splitting on just the first occurrence might not work if first `.` is the wrong one. I'll add validation to crash the process if something is unexpected

workspace/notebook/py_etl_executor.json
HarrisonBoyleThomas: I'll update this

workspace/notebook/py_etl_orchestrator.json
HarrisonBoyleThomas: Added in

odw/core/orchestration/orchestration_config.py
HarrisonBoyleThomas: Updated

Fred83200 (approved):

PR #2540, opened 2026-06-04
Fix/theodw 3217 optimise nsip project

HarrisonBoyleThomas:
Ticket: https://pins-ds.atlassian.net/browse/THEODW-3217

Replace inefficient aggregations in the NSIP project ETLProcess (inherited from the source notebook) with a window function - cutting down test runtime from ~10m to ~1m

### DIFF
--- a/odw/core/etl/transformation/harmonised/nsip_project_harmonisation_process.py
+++ b/odw/core/etl/transformation/harmonised/nsip_project_harmonisation_process.py
@@ -881,7 +881,7 @@ class NsipProjectHarmonisationProcess(HarmonisationProcess):
         grouping_cols_map = {"region": "regions"} | {x: x for x in grouping_cols}
         # This deviates from the original notebook, which has performance issues
         w = Window.partitionBy(horizon_data["caseid"], "IngestionDate")
-        horizon_data = horizon_data.withColumns({alias: F.collect_list(col).over(w).alias(alias) for col, alias in grouping_cols_map.items()})
+        horizon_data = horizon_data.withColumns({alias: F.collect_list(col).over(w) for col, alias in grouping_cols_map.items()})
         # Sort columns into same order as service bus
         horizon_data = horizon_data.select(service_bus_data.columns)
 
--- a/odw/core/etl/transformation/harmonised/nsip_project_harmonisation_process.py
+++ b/odw/core/etl/transformation/harmonised/nsip_project_harmonisation_process.py
@@ -498,10 +498,11 @@ class NsipProjectHarmonisationProcess(HarmonisationProcess):
         """
         # Need to read the original table
         w_reverse_per_case = Window.partitionBy("caseid").orderBy(F.col("IngestionDate").desc())
+        w_internal_id = Window.orderBy(F.col("IngestionDate").asc(), F.col("caseid").asc())
 
         out_df = combined_data.select(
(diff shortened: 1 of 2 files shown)


odw/core/etl/transformation/harmonised/nsip_project_harmonisation_process.py
```diff
@@ -881,10 +879,9 @@ def process(self, **kwargs):
...
         grouping_cols_map = {"region": "regions"} | {x: x for x in grouping_cols}
-        for col, alias in grouping_cols_map.items():
-            sub_grouping = horizon_data_grouped.agg(F.collect_list(col).alias(alias))
-            horizon_data = horizon_data.drop(col).join(sub_grouping, on=["caseid", "IngestionDate"], how="inner")
-
+        # This deviates from the original notebook, which has performance issues
+        w = Window.partitionBy(horizon_data["caseid"], "IngestionDate")
+        horizon_data = horizon_data.withColumns({alias: F.collect_list(col).over(w).alias(alias) for col, alias in grouping_cols_map.items()})
         # Sort columns into same order as service bus
         horizon_data = horizon_data.select(service_bus_data.columns)
```
stef-solirius, line 886: I might be wrong, but shouldn't we add add .distinct() after the select?

The collect_list doesn't deduplicate. When horizon data has multiple rows for the same (caseid, IngestionDate) the new code keeps all of them instead of collapsing to one per group. After select(service_bus_data.columns) they become identical, but _clean_data's row_number() then assigns them different IsActive values, so drop_duplicates() at the end can't remove them. The output ends up with duplicate case records.

also, the existing tests don't catch this because test__aggregates_regions_using_collect_list and test__aggregates_horizon_array_like_id_fields both use .collect()[0] - they check the content of the first row but never assert that only one row was returned.

odw/core/etl/transformation/harmonised/nsip_project_harmonisation_process.py
```diff
@@ -881,10 +879,9 @@ def process(self, **kwargs):
             "environmentalServicesOfficerIds",
         )
         grouping_cols_map = {"region": "regions"} | {x: x for x in grouping_cols}
-        for col, alias in grouping_cols_map.items():
-            sub_grouping = horizon_data_grouped.agg(F.collect_list(col).alias(alias))
-            horizon_data = horizon_data.drop(col).join(sub_grouping, on=["caseid", "IngestionDate"], how="inner")
-
+        # This deviates from the original notebook, which has performance issues
+        w = Window.partitionBy(horizon_data["caseid"], "IngestionDate")
+        horizon_data = horizon_data.withColumns({alias: F.collect_list(col).over(w).alias(alias) for col, alias in grouping_cols_map.items()})
```
stef-solirius: The dict key already sets the output column name so the .alias(alias) on the expression is a no-op. Fine to drop it.

HarrisonBoyleThomas: Oh yeah good spot

odw/core/etl/transformation/harmonised/nsip_project_harmonisation_process.py
HarrisonBoyleThomas: Lemme update the test case for this

stef-solirius (approved):

PR #2545, opened 2026-06-05
feat: THEODW-3251 unnest events JSON columns in nsip exam timetable harmonised and curated processes

stef-solirius:
## Summary

- Replaces the `events` JSON array column with individual flat columns (`eventId`, `type`, `eventTitle`, `eventTitleWelsh`, `description`, `descriptionWelsh`, `eventDate`, `eventDeadlineStartDate`) in both the harmonised and curated NSIP Exam Timetable processes
- Service Bus records: explodes the `events` struct array into one row per event, applying empty-string → null coercion and date formatting (`yyyy-MM-dd HH:mm`)
- Horizon records: SQL updated to produce `eventDate` / `eventDeadlineStartDate` as pre-formatted strings rather than raw datetime columns; events stay as individual rows (no more `collect_list` aggregation)
- RowID hash updated from `MD5(CONCAT(...))` to `MD5(CONCAT_WS('|', ...))` over the full set of individual event columns including `SourceSystemID`
- Step 7 rejoin in the harmonised process changed from `(caseReference, IngestionDate)` key to `NSIPExaminationTimetableID` to avoid an N×N cartesian product when multiple events share the same ingestion date
- Curated `load_data` and `process` updated to select and expose the ten individual event columns instead of the `events` struct

## Test plan

- [ ] Unit test harmonised: `test__nsip_exam_timetable_harmonisation_process__process__explodes_events_and_sets_published` — verifies explode, empty-string → null, Migrated flag, and SourceSystemID in output
- [ ] Unit test curated: `test__nsip_exam_timetable_curated_process__process__keeps_only_projects_in_curated_project_table` — verifies Horizon-only filter, latest-ingestion logic, and individual column output
- [ ] Integration test harmonised: end-to-end run against real Delta tables, checks exploded row count and column values
- [ ] Integration test curated: end-to-end run with absolute-path fix for Parquet metastore consistency
- [ ] `ruff check` passes on all changed files


odw/test/integration_test/etl/curated/test_nsip_exam_timetable_curated_process.py
```diff
@@ -1,11 +1,16 @@
...
 from odw.test.util.session_util import PytestSparkSessionUtil
 from odw.test.util.assertion import assert_etl_result_successful
-from pyspark.sql import Row
 import pyspark.sql.types as T
 import mock
+import os
+
+
+def _abs_format_to_adls_path(inst, container_name: str, blob_path: str, **kwargs) -> str:
+    return os.path.abspath(os.path.join(container_name, blob_path))
```
HarrisonBoyleThomas: You shouldn't need to do this - `odw.test.util.util` has functions that should be used to patch `format_to_adls_path`

odw/test/integration_test/etl/curated/test_nsip_exam_timetable_curated_process.py
```diff
@@ -62,6 +63,7 @@ def test__nsip_exam_timetable_curated_process__run__keeps_only_projects_in_curat
             spark, curated_projects, curated_projects_table, "odw_curated_db", "odw-curated", curated_projects_table, "overwrite"
         )
         output_table = f"{test_case}_nsip_exam_timetable"
+        spark.sql(f"DROP TABLE IF EXISTS odw_curated_db.{output_table}")
```
HarrisonBoyleThomas: You shouldn't need to do this, the DB is cleared before the test begins. If there's a discrepancy it may be due to a collision with the `test_case` variable

odw/test/integration_test/etl/curated/test_nsip_exam_timetable_curated_process.py
```diff
@@ -71,6 +73,7 @@ def test__nsip_exam_timetable_curated_process__run__keeps_only_projects_in_curat
             mock.patch.object(NsipExamTimetableCuratedProcess, "HARMONISED_TABLE", f"odw_harmonised_db.{harmonised_exam_timetable_table}"),
             mock.patch.object(NsipExamTimetableCuratedProcess, "CURATED_PROJECT_TABLE", f"odw_curated_db.{curated_projects_table}"),
             mock.patch.object(NsipExamTimetableCuratedProcess, "OUTPUT_TABLE", output_table),
+            mock.patch.object(SynapseTableDataIO, "_format_to_adls_path", _abs_format_to_adls_path),
```
HarrisonBoyleThomas: You shouldn't need to patch this - it should be done in the parent class (and if it's not then it should be done there imo)

odw/test/integration_test/etl/curated/test_nsip_exam_timetable_curated_process.py
stef-solirius: Replaced, thanks

odw/test/integration_test/etl/curated/test_nsip_exam_timetable_curated_process.py
stef-solirius: Removed, thanks

odw/test/integration_test/etl/curated/test_nsip_exam_timetable_curated_process.py
stef-solirius: Removed, thanks

odw/core/etl/transformation/harmonised/nsip_exam_timetable_harmonisation_process.py
```diff
@@ -157,7 +143,42 @@ def process(self, **kwargs) -> Tuple[Dict[str, DataFrame], ETLResult]:
         horizon_nsip_data: DataFrame = self.load_parameter("horizon_nsip_data", source_data)
         sb_case_references: DataFrame = self.load_parameter("sb_case_references", source_data)
 
-        # Step 1: Join Horizon with horizon_nsip_data and align to SB schema
+        # Step 1: Explode service bus events into individual rows
+        LoggingUtil().log_info("Exploding service bus events into individual rows")
+        service_bus_event_data = service_bus_data.select(
```
HarrisonBoyleThomas: I think this should be squashed into a single select - i've also raised this in the other PR

stef-solirius: Done, thank you.

HarrisonBoyleThomas (approved):

PR #2566, opened 2026-06-18
fix: correct Purview qualified name for ServiceBus entities with digits

stef-solirius:
## Summary

- Purview keeps the entity folder name literal but replaces digit sequences in the filename prefix with `{N}` — the code was applying `{N}` substitution to both, causing a 404 for entities like `appeal-s78` and `s51-advice`
- Time component placeholder was wrong: `T{Hour}_{Minute}_{Second}` (Purview has no named placeholder for Minute/Second, and files use ISO 8601 colons) — corrected to `T{Hour}:{N}:{N}`
- Added `_PurviewEntityNotFoundError` so a 404 from Purview (uncatalogued asset) returns an empty classification list rather than failing the standardisation job

## Test plan

- [ ] `poetry run pytest odw/test/unit_test/anonymisation/test_unit_anonymisation_engine.py` — all 27 unit tests pass
- [ ] Exact-match FQN tests added for `appeal-s78` and `s51-advice`
- [ ] Re-run `py_sb_raw_to_std` in DEV for `appeal-s78` and `s51-advice` to confirm no more 404


odw/core/anonymisation/engine.py
```diff
@@ -227,8 +228,14 @@ def _purview_base_url(purview_name: str) -> str:
     return f"https://{purview_name}.catalog.purview.azure.com/api/atlas/v2"
 
 
+class _PurviewEntityNotFoundError(Exception):
```
HarrisonBoyleThomas: I'd remove the `_` prefix since this is being publicly thrown

HarrisonBoyleThomas (approved):

PR #2568, opened 2026-06-18
Feat/2410 implement logging and tracking in pln_copy_entraid_curated_mipins

PraveenKomariPINS:
**Ticket Reference:**
https://pins-ds.atlassian.net/browse/THEODW-2410?assignee=712020%3A0281513f-436b-47bc-b170-d1962f8c9715&search_id=f30f4e5b-f54a-4ce0-96c2-8a6e419f21bf

**Description:**

I have implemented logging and monitoring in the pln_copy_entraid_curated_mipins pipeline.
For each pipeline execution, I now log a single entity containing:
• Pipeline start time
• Pipeline end time
• Total duration
• Pipeline status (Success / Failed)
• Error messages (when tests fail)

**Implementation details:**
• Added start and completion Execute Pipeline calls to pln_log_to_appins
  (logging pipeline name, run id, start/end time, status, and trigger metadata).
• Cleaned up the copy activity (removed SELECT 1/0 test script) and added variables
  to capture pipeline status and any error text.
• Updated Sp_log_success so ErrorMessage is null on success and populated only when
  the copy fails.
• Ensured an IfCondition + Fail activity is used so the pipeline explicitly fails
  when any step fails, avoiding silent failures.

**Application Insights:**
• Verified that pln_copy_entraid_curated_mipins now logs start and completion events
  to Application Insights via pln_log_to_appins, with key fields (PipelineName,
  PipelineRunId, Stage, StatusMessage, ErrorMessage, durationSeconds) matching the
  actual execution.
  
  
image


workspace/notebook/Entraid_curated_mipins.json
```diff
@@ -51,67 +51,153 @@
...
+			{
+				"cell_type": "code",
+				"source": [
+					"PipelineName = \"pln_curated_mipins\"\r\n",
+					"PipelineRunID = \"097e997a-70d4-4188-90c8-09b9c7445c8e\"\r\n",
+					"PipelineTriggerID = \"e72971fa-9f51-4683-98c0-d7d93425023f\"\r\n",
+					"PipelineTriggerName = \"e72971fa-9f51-4683-98c0-d7d93425023f\"\r\n",
+					"PipelineTriggerType = \"PipelineActivity\"\r\n",
+					"PipelineTriggeredbyPipelineName = \"pln_master\"\r\n",
+					"PipelineTriggeredbyPipelineRunID = \"688cd58a-1365-4e39-8d5e-dd39745c15c5\""
```
HarrisonBoyleThomas: You might want to set some of these to `None`, to catch any cases where this is accidentally not provided when calling the notebook

HarrisonBoyleThomas: Or maybe set it to a new uuid4 by default, just to avoid collisions

PraveenKomariPINS: I’ve removed the hard‑coded values from this cell so we don’t risk collisions or stale data, and the pipeline now passes these in when it calls the notebook

KranthiRayipudi (approved):

PR #2571, opened 2026-06-19
Feat/332 service bus loadfailing

KranthiRayipudi:
[https://pins-ds.atlassian.net/browse/THEODW-3323](url)

Added retry logic around the MERGE operation
Handled ConcurrentAppendException to avoid failures
Added delay (sleep) before retrying


HarrisonBoyleThomas (commented):
I think this solution kicks the problem down the road and will slow the runtime down. The notebook already has a retry mechanism in place, ideally we can just prevent the error in the first place. https://docs.delta.io/concurrency-control/ has some suggestions

KranthiRayipudi:
Thanks for the suggestion. I’ve removed the retry logic and replaced the MERGE with a partition-level overwrite (replaceWhere). Since we only store one record per key, this avoids concurrency conflicts and simplifies the implementation

HarrisonBoyleThomas (commented):
Hm i think you could achieve this with a simpler approach. You could modify the existing merge statement to tighten the `on` condition to include the `run_id`, which should guarantee uniqueness within a pipeline run

If collisions still occur, then we will need to investigate why, since at that stage it means that two processes are writing entries in the same partition, which shouldn't be happening

KranthiRayipudi:
Thanks for the suggestion.
I’ve updated the implementation by the MERGE condition to include run_id. This ensures uniqueness within each pipeline run and reduces the likelihood of conflicts when writing to the same partition.

HarrisonBoyleThomas (approved):

PR #2574, opened 2026-06-22
Theodw 2434 implementing logging and tracking in entra

anushaveerapalli68:
JIRA Ticket Reference: https://pins-ds.atlassian.net/browse/THEODW-2434

Error Handling:
Enhanced error handling by introducing structured error messages (including step name, exception type, and message), with fallback logic to handle formatting failures, limiting messages to 800 characters, wrapping logError in try/catch with print fallback, and safely integrating optional telemetry via _log_telemetry without impacting execution.
Success Update:
Updated stage value to "Success" where applicable for accurate execution tracking.
Set Variable:
Added a Set Variable activity at the beginning to capture pipeline start time.


KranthiRayipudi:
1.You can remove the "Record Horizon EntraID Data" activity. Since we are planning to maintain a single entry per pipeline execution, there is no need for a separate start record activity. Instead, capture the start time in a variable at the beginning of the pipeline and use that variable in the "Record Completed EntraID Data" activity.

2.The DurationSeconds calculation is currently missing. Please calculate it using the stored start-time variable and the current end time (@utcNow()). The duration should represent the total execution time of the pipeline in seconds.

3.The StatusCode field is also missing. Set the StatusCode to:
200 when the pipeline execution completes successfully.
500 when an error message is generated or the pipeline execution fails.

4.After implementing these changes, please execute the pipeline and provide the results in an Excel file, including all Application Insights logging results for validation.
image

anushaveerapalli68:
> 1.You can remove the "Record Horizon EntraID Data" activity. Since we are planning to maintain a single entry per pipeline execution, there is no need for a separate start record activity. Instead, capture the start time in a variable at the beginning of the pipeline and use that variable in the "Record Completed EntraID Data" activity.
> 
> 2.The DurationSeconds calculation is currently missing. Please calculate it using the stored start-time variable and the current end time (@utcnow()). The duration should represent the total execution time of the pipeline in seconds.
> 
> 3.The StatusCode field is also missing. Set the StatusCode to: 200 when the pipeline execution completes successfully. 500 when an error message is generated or the pipeline execution fails.
> 
> 4.After implementing these changes, please execute the pipeline and provide the results in an Excel file, including all Application Insights logging results for validation. image

- Removed "Record Horizon EntraID Data" activity to ensure single logging per pipeline execution
- Captured pipeline start time using variable start_time = @utcnow() at pipeline beginning
- Implemented DurationSeconds and PipelineExecutionTimeInSec using start time and @utcnow()
- Added StatusCode logic (200 for success, 500 for failure) based on activity error aggregation
- Implemented Stage field to return Success / Failed based on pipeline execution
- Implemented StatusMessage for meaningful success/failure description
- Implemented ErrorMessage aggregation across Src → Raw → Std → Hrm activities
- Configured StartTime and EndTime for full execution tracking
- Set default values (0) for Inserts, Updates, Deletes
- Used pipeline-level metadata (PipelineName, RunId, Trigger details) instead of activity-level values
- Configured Application Insights logging with event name ODW_Master_Pipeline_Logs
- Removed all references to deleted activity to resolve validation errors
- Ensured logging activity depends on all upstream activities to avoid ancestor issues

image

KranthiRayipudi (approved):

PR #2591, opened 2026-06-30
THEDOW-2412: Implement logging and tracking for appeal_event_estimate

PraveenKomariPINS:
**Ticket Reference:** 
https://pins-ds.atlassian.net/browse/THEODW-2412?search_id=493e1bf2-1095-43a8-b8f7-d21ad0ca6576

**Descrption:**

Summary
- Updated notebook appeal_event_estimate_curated_mipins to follow the standard telemetry pattern used by other appeal notebooks.
- Ensured the notebook logs start/end times, status, row counts and error messages via log_telemetry_and_exit.
- Updated pipeline pln_copy_appeal_event_estimate_curated_mipins to use the notebook and to surface success/failure in Application Insights (ODW_Master_Pipeline_Logs).

Testing
- Ran the notebook manually: completed successfully end-to-end with no NameError or telemetry issues.
- Ran the pipeline with a forced failure to verify error logging; confirmed failed entries appear in ODW_Master_Pipeline_Logs for pln_copy_appeal_event_estimate_curated_mipins.
- Restored the dataset to point back to the real curated table and re-ran the pipeline.
- Ran the pipeline successfully; confirmed “Successfully copied data…” entry in App Insights with correct Stage, StatusMessage, StartTime, EndTime and PipelineRunID.

image


anushaveerapalli68:
1. You can remove the "Record start appeal event" activity. Since we are planning to maintain a single entry per pipeline execution, there is no need for a separate start record activity. Instead, capture the start time in a variable at the beginning of the pipeline and use that variable in the "Record completed appeal event estimate curated mipins" activity.
2. The Stage value should be updated to "Success" where applicable instead of "Successfully" to maintain consistency with the existing telemetry logging standards (Success / Failed).

KranthiRayipudi:
The Pass parameter value is missing from the pln_curated_mipins main pipeline and is not being passed to the appeal_event_estimate_curated_mipins activity notebook .

KranthiRayipudi (approved):

PR #2600, opened 2026-07-06
Feat/theodw 2996 code coverage

HarrisonBoyleThomas:
Ticket: https://pins-ds.atlassian.net/browse/THEODW-2996

- Update ODW package test cases to fail if coverage is below a defined threshold (different threshold for each category)
- Update CI pipeline to fail if overall ODW package test coverage is below a global threshold
- Update CI pipeline to publish code coverage for the ODW package tests


KalyaniNik (approved):
LGTM

PR #2607, opened 2026-07-08
add logic for parsing status flag

NHAL22:
1. JIRA Ticket Reference  :
https://pins-ds.atlassian.net/browse/THEODW-3380

 2.  Summary of the work : 
- A `parse_status_flag` helper that safely converts the flag value from the orchestration.json to a python boolean, guarding against type inconsistencies from spark's JSON reader (example, the string `"false"` being misread as truthy by plan `bool()`)
- The existing `get_table_created_status` function's final return was updated to use `parse_status_flag` instead of `bool()`. This function reads orchestion.json, looks up the entity by `source_filename_start`, and returns whether curated table creation is permitted. If the flag is absent (all exisiting entities) it will default to `True` - no behaviour change for those entities.
- When `get_table_created_status` returns False, the notebook logs a telemetry event and exits cleanly via mssparkutils.notebook.exit('') which the pipeline treats as **Succeeded**, allowing Raw to Std and Std to Hrm to continue normally for application-update


workspace/notebook/create_table_from_schema.json
```diff
@@ -211,6 +211,91 @@
...
+					"    status_field_map = {\n",
+					"        \"odw_curated_db\": \"odw_curated_db_table_status\",\n",
+					"        \"odw_curated_migration_db\": \"odw_curated_db_table_status\",\n",
+					"        \"odw_standardised_db\": \"odw_standardised_db_table_status\",\n",
+					"        \"odw_harmonised_db\": \"odw_harmonised_db_table_status\"\n",
+					"    }\n",
+					"\n",
+					"    status_field = status_field_map.get(db_name)\n",
+					"    if not status_field:\n",
+					"        return True\n",
```
HarrisonBoyleThomas: This could be removed, unless you extend the `.get` to have a `default=None`

workspace/notebook/create_table_from_schema.json
```diff
@@ -211,6 +211,91 @@
...
 				"source": [
+					"def parse_status_flag(value) -> bool:\n",
+					"    if isinstance(value, bool):\n",
+					"        return value\n",
+					"    if isinstance(value, str):\n",
+					"        return value.strip().lower() in (\"true\", \"1\", \"yes\", \"y\")\n",
+					"    if isinstance(value, (int, float)):\n",
+					"        return value != 0\n",
+					"    return False\n",
+					"\n",
```
HarrisonBoyleThomas: Since this seems to be a new property in the config, i'd simplify this to just expect it to be a boolean

workspace/notebook/create_table_from_schema.json
```diff
@@ -211,6 +211,91 @@
...
+					"@logging_to_appins\n",
+					"def get_table_created_status(db_name: str, entity_name: str, path_to_orchestration_file: str) -> bool:\n",
```
HarrisonBoyleThomas: Similar to my comment at https://github.com/Planning-Inspectorate/odw-config/pull/186 , i'd give this a more descriptive name

workspace/notebook/create_table_from_schema.json
NHAL22: I thought there maybe a risk it could come through as a string "false" or some other type and bool("false") in python returns True (any non-empty string being truthy), which would be wrong but since its only in one place and when spark reads a JSON boolean it serialises it back (via toJSON()) and produces a json string with false/true which json.loads() converts it back to a Python bool so the type is preserved throughout? @HarrisonBoyleThomas

HarrisonBoyleThomas: Hm true, i guess if there's a misconfiguration. Maybe it might be worth checking that it's a string, and if it's not then raising an exception? Separately we should add validators for the ODW config

HarrisonBoyleThomas (dismissed):

workspace/notebook/create_table_from_schema.json
```diff
@@ -130,8 +130,8 @@
 					]
 				},
 				"source": [
-					"db_name: str = 'odw_curated_db'\n",
-					"entity_name: str = 'application-update'\n",
+					"db_name: str = ''\n",
+					"entity_name: str = ''\n",
```
HarrisonBoyleThomas, line 134: I'd keep this as-is incase there are default values being used

HarrisonBoyleThomas: Ignore this, see it was from a previous commit

HarrisonBoyleThomas (approved):

PR #2610, opened 2026-07-09
Feat/3353 implementin logging and monitoring inpln master

KranthiRayipudi:
[https://pins-ds.atlassian.net/browse/THEODW-3353?actionerId=712020%3A998f498a-de2b-4088-9fc8-4433bf0af776&sourceType=assign&atlOrigin=eyJpIjoiNTQ0YzEyNDBhZWZkNGE1N2FlMGM4YzMxYzdjMDgxMDAiLCJwIjoiaiJ9](url)

I have removed the following trace activities from the pln_master pipeline because tracing is now being implemented within the relevant sub-pipelines. This approach ensures that only a single trace entry is generated for each pipeline in Application Insights, helping to avoid duplicate logging and improving traceability.
Removed Trace Activities

Record pipeline starting
Record moving to loading Horizon data
Record moving to loading Service Bus data
Record moving to loading Curated data
Record pipeline completed
Record Horizon failure
Record Curated failure
Record Service Bus failure

In addition, I have added a new activity, "Record pln_master Pipeline Completed", immediately before the execution of the py_logging_from_app_insights notebook. This ensures that the completion trace is written to Application Insights before the logging notebook is executed.
This change addresses an issue where some trace records were not appearing in the Power BI report. Since failures can now be traced directly at the sub-pipeline level, and the main pipeline already captures sub-pipeline execution outcomes, the additional trace activities in pln_master are no longer required.
The primary purpose of retaining tracing at the pln_master level is to capture the overall pipeline execution metrics, including:

Master pipeline start time
Master pipeline end time
Total execution duration
Overall pipeline status

The pipeline will continue to return a 200 status code upon successful completion. Any failure details can be investigated through the corresponding sub-pipeline traces recorded in Application Insights.
Note
The pln_master pipeline status can still be monitored through the Power BI report. Based on the master pipeline start and end times, we can identify all related execution events that occurred during that period. If any error messages are recorded in Application Insights within the execution window, the report can display a Warning or Failed status, indicating that a sub-pipeline or notebook encountered an issue during the run.
This approach allows the report to accurately identify and display the specific failure message and the component (sub-pipeline or notebook) responsible for the failure, while keeping tracing centralized and avoiding duplicate records.
Please find attached the results from the Application Insights test run and the corresponding validation results.
[pln_master.csv](https://github.com/user-attachments/files/29851187/pln_master.csv)


workspace/pipeline/pln_master.json
```diff
@@ -1773,11 +1221,14 @@
 			},
 			"H_NoteBook_errorMessage": {
 				"type": "Array"
+			},
+			"pln_master_ Status": {
```
KalyaniNik: pln_master_ Status,  space before "Status" in the variable name. Also I think this variable is not required as I don't see its being set anywhere . Could you please check

KranthiRayipudi: I have removed unused variable

KalyaniNik (approved):
LGTM

PR #2628, opened 2026-07-16
Feat/3393 py horizon raw to std notebook

KranthiRayipudi:
As part of the Application Insights logging enhancement, the notebook execution flow was updated to centralize telemetry logging using the log_telemetry_and_exit() function. The implementation continues to log both success and failure statuses of the py_horizon_raw_to_std notebook to Application Insights. Error handling was improved by adding explicit failure logging while ensuring that telemetry is always captured. The notebook behaviour remains aligned with the original requirement, allowing the pipeline to continue executing subsequent activities regardless of the notebook outcome, while providing better visibility into failures for future review and remediation.


workspace/notebook/py_horizon_raw_to_std.json
anushaveerapalli68, line 1: Thanks for the changes. The implementation looks aligned with the requirement to capture notebook execution status in Application Insights.

Before approving, I'd like to see the validation against the acceptance criteria from the ticket:
1. Confirmation that log_telemetry_and_exit() does not interrupt execution and allows processing to continue as expected.
2. Test results showing a failure scenario where one file fails and the remaining files continue to be processed.
3. Row count and final output validation before and after the change, as requested in the ticket.

If these scenarios have already been tested, please attach the test evidence/results so we can confirm the expected behaviour and complete the review.

KranthiRayipudi: [pln_horizon_to_odw.csv](https://github.com/user-attachments/files/30090075/pln_horizon_to_odw.csv)
I passed three files for processing at the same time. One file failed, while the other two loaded successfully. Application Insights confirms this behavior, showing one failed file and the remaining files completed successfully.
image

anushaveerapalli68: Thanks for sharing the test results. I can see that when AppealsAdditionalData.csv failed, the other files were still processed successfully, which demonstrates the required continuation behaviour.

1. App Insights entries

I noticed multiple entries under the same PipelineRunID such as:

image

Successfully loaded data from a into None table
Successfully loaded data from b into None table
Successfully loaded data from c into None table
Successfully loaded data from d into None table

Some of these entries appear more than once within the same run. If these are being treated as individual entities/files, can you clarify why they are being processed and logged multiple times after a successful execution? I'm trying to understand whether this is expected behaviour, a retry scenario, or part of the telemetry testing used for validation.

2. Row count and final output validation

The ticket specifically mentions validating both the row count and final output before and after the change. I don't think I've seen that evidence yet. Could you please share the row count comparison and final output validation results as well?

3. log_telemetry_and_exit() behaviour

Since this method is central to the solution, can you confirm that log_telemetry_and_exit() only handles telemetry logging and does not internally call mssparkutils.notebook.exit() or throw exceptions that could interrupt notebook execution? The test results suggest processing continues as expected, but it would be good to confirm the helper method's behaviour as well.

KranthiRayipudi: Thanks for review, I have tested the changes by running the notebook multiple times and creating failure scenarios. For the initial validation, I used three test files, with one of the files intentionally configured to fail. As shown in the attached telemetry logs, the remaining files continued to process successfully after the failure occurred, demonstrating that execution is not stopping due to the telemetry logging changes.
           I have already provided these test results in the ticket and the PR. The purpose of this change was to validate the notebook's behaviour using test data and telemetry logs, not provided whatever I have tried initial logs .
There are approximately 30 files in total, and it is not practical to test all 30 files individually for this change. Please let me know how many files you would like me to validate as part of the testing.
            If you believe there is a code issue, could you please provide specific evidence or an example from the logs that indicates a problem? Based on the test results and telemetry data shared, I have not found any evidence that this code change causes notebook execution to stop or impacts the processing of subsequent files. The behaviour observed is consistent with the requirements of this ticket.
         Currently, the ODW Logging and Monitoring implementation is used solely for writing telemetry information to Application Insights. I can confirm that log_telemetry_and_exit() does not call mssparkutils.notebook.exit() anywhere, nor does it contain any logic that stops, terminates, or exits notebook execution. Its purpose is limited to telemetry logging and monitoring.

        I am happy to change code  if suggest code problem and suggestion related this ticket.

anushaveerapalli68: Thanks for the detailed explanation. I agree that the test results demonstrate that processing continues when an entity fails and that `log_telemetry_and_exit()` does not terminate notebook execution.

My question is more about the monitoring behaviour introduced by this change rather than the continuation logic itself.

Looking at the current implementation, telemetry is being logged at the individual entity/file level:

stage = "Success" if not error_message else "Failed"
log_telemetry_and_exit(...)

and when an entity fails, the code logs the error and continues processing the remaining entities.

I don't see any notebook-level aggregation of failures or any logic that marks the overall notebook execution as Failed when one or more entities fail.

For example:

42 entities processed
41 Success
1 Failed

In this scenario, the failed entity would be logged to Application Insights, the remaining entities would continue processing, and the notebook would still reach:

logInfo("Completed processing all Horizon files.")

With the previous implementation, a single entity failure would result in the notebook execution being marked as Failed. With the current implementation, am I correct in understanding that the notebook execution itself will now complete successfully, with failures only visible at the individual entity level in Application Insights?

If that's the intended behaviour, how are partial failure scenarios (for example, 41 successful entities and 1 failed entity) expected to be identified from a monitoring and operational support perspective?

Also, the ticket specifically mentions validating row counts and final output before and after the change. Could you please share the row count comparison and final output validation results as well?

KranthiRayipudi: Regarding the monitoring aspect, the current implementation continues to provide visibility of individual file/entity failures through Application Insights, which is the same monitoring approach used previously. Each file is tracked independently, and when a specific file fails, a telemetry record is logged with Stage = Failed, along with the corresponding error message and status code.
As shown in the attached screenshot, the failed file is clearly visible in Application Insights, including the detailed error information. These failures are also surfaced in the Power BI monitoring dashboard, where failed files are highlighted in red, making partial failure scenarios straightforward to identify and investigate. The associated error message can be viewed directly from the telemetry records for troubleshooting purposes.
Therefore, in a scenario where 41 entities are processed successfully and 1 entity fails, the failed entity remains visible in both Application Insights and the monitoring dashboard, while processing continues for the remaining entities. This ensures that a single file failure does not prevent other files from being loaded.
In addition, the notebook execution reaches completion only after the for loop has finished processing all files in horizon_files, regardless of whether individual files succeeded or failed. This behaviour is intentional and enables processing of all available files within the same run.
Regarding the row count and output validation, please find the attached comparison results. The before-and-after row counts were validated against both the source files and target tables, and the results are consistent with the expected output. No discrepancies were identified as part of the validation exercise.
image
image

anushaveerapalli68: Thanks for addressing all the review comments and providing the additional clarification, test evidence, and validation results. The explanation around the monitoring approach, entity-level failure tracking in Application Insights/Power BI, and the row count/output validation answers my concerns.

prathapA100 (approved):
LGTM

PR #2629, opened 2026-07-16
Updating notebook: py_comment_type_ref

prathapA100:
Comment type reference - table is empty because name has given incorrect 

image

image


anushaveerapalli68 (approved):
LGTM

PR #2638, opened 2026-07-20
task/dev-772-package-deployment

beejjacobs:
https://pins-ds.atlassian.net/browse/DEV-772

Consolidate package deployment, combining the deploy_odw_package script and the deploy_packages script from the infrastructure repo.


pipelines/scripts/packages/spark_pool_packages.py
```diff
@@ -0,0 +1,35 @@
...
+    name: str
+    type: str
+    properties: LibraryInfo
+
+class WorkspacePackages:
+    """
+    Package information for a Synapse workspace
+    """
+    def __init__(self):
+        pass
```
HarrisonBoyleThomas: This can be removed

pipelines/scripts/packages/spark_pool_packages.py
```diff
@@ -0,0 +1,32 @@
...
+    properties: LibraryInfo
+
+class WorkspacePackages:
+    """
+    Package information for a Synapse workspace
+    """
+    packages: List[LibrariesResponseItem]
+    packageNames: Set[str]
+    odw: List[LibrariesResponseItem]
+    odwNames: Set[str]
```
HarrisonBoyleThomas: I would use snake_case here for consistency

pipelines/scripts/test/packages/test_spark_pool_configuration.py
```diff
@@ -0,0 +1,149 @@
+from typing import List
+
+from pipelines.scripts.packages.spark_pool_configuration import SparkPoolConfiguration
+from pipelines.scripts.packages.synapse_types import BigDataPoolResourceInfo, LibraryInfo
+
+
+ENV = "test"
```
HarrisonBoyleThomas, line 7: Might need to load this in using `dotenv` so it's not hardcoded

pipelines/scripts/test/packages/test_spark_pool_packages.py
```diff
@@ -0,0 +1,120 @@
...
+
+def test_partitions_odw_and_non_odw_packages():
+    packages = [
+        {"name": "odw-1.0.0-abc-py3-none-any.whl"},
+        {"name": "mypackage.whl"},
+        {"name": "odw-1.0.0-def-py3-none-any.whl"},
+        {"name": "otherlib.jar"},
+    ]
+    with mock.patch(WORKSPACE_MANAGER_PATH) as mock_workspace_manager:
+        mock_workspace_manager.get_workspace_packages.return_value = packages
```
HarrisonBoyleThomas: With this you should be able to do 
```
with mock.patch.object(SynapseWorkspaceManager, "get_workspace_packages", return_value=packages):
```
instead of patching the specific path to it, although this is fine and works

pipelines/scripts/test/packages/test_spark_pool_packages.py
```diff
@@ -0,0 +1,120 @@
+import mock
```
HarrisonBoyleThomas, line 1: I think it'd be good to add a test to cover the below case

- Given I have an odw package deployed to an "unknown" spark pool (e.g a pool called mySparkPool not included in the infra/the config here)
- When I deploy a new version of the odw package
- Then: the old odw package should be unbound from only the known spark pools, with the new odw package uploaded/attached only to the known spark pools. The old odw package should be left in the workspace, and only attached to the "unknown" spark pool

This is the functionality that currently exists with the older approach

pipelines/scripts/test/packages/test_spark_pool_configuration.py
beejjacobs: Only for unit tests, does this matter?

pipelines/scripts/test/packages/test_spark_pool_packages.py
beejjacobs: Tests added

HarrisonBoyleThomas (approved):

PR #2639, opened 2026-07-20
Feat/theodw 1893 historical data anonymisation

HarrisonBoyleThomas:
Ticket: https://pins-ds.atlassian.net/browse/THEODW-1893

- Move the `mock` module to `mock_util` to avoid collisions with the python `mock` library (update all imports to reflect the change, which is why so many files were modified)
- Fix imports in the anonymisation engine
- Update SynapseFileDataIO to allow multiple specific files to be read (for loose csv files in different folders)
- Add ETLProcesses to handle historical anonymisation
- Add unit/integration tests for the new ETLProcess
- Add a pipeline to trigger the historical anonymisation across the relevant entities
- Update pyproject to have a `authors` block to allow local builds


workspace/pipeline/pln_historical_anonymisation.json
```diff
@@ -0,0 +1,126 @@
...
+				"dependsOn": [],
+				"policy": {
+					"secureOutput": false,
+					"secureInput": false
+				},
+				"userProperties": [],
+				"typeProperties": {
+					"variableName": "entities",
+					"value": {
+						"value": "@createArray(\n    'appeal-has',\n    'nsip-subscription',\n    'InspectorCases',\n    'DaRT_Inspectors',\n    'CaseInvolvement',\n    'DaRT_LPA',\n    'HorizonCases_s78',\n    'NSIPReleventRepresentation',\n    'HorizonCases_Has',\n    'DocumentMetaData',\n    'CaseSiteStrings'\n)",
```
KalyaniNik, line 71: Do we need to include AIEDocumentData

odw/core/io/synapse_file_data_io.py
```diff
@@ -83,18 +85,35 @@ def read(self, **kwargs) -> DataFrame:
...
-            )
+        if isinstance(blob_path, str):
+            blob_paths = [blob_path]
+        elif isinstance(blob_path, list):
+            blob_paths = blob_path
         else:
-            data_path = self._format_to_adls_path(
-                container_name, blob_path, storage_endpoint=storage_endpoint
+            raise ValueError(
+                f"Expected blob_path to be a list or a string, but was a {type(blob_paths)}"
```
KalyaniNik: it should be blob_path to blob_paths

workspace/pipeline/pln_historical_anonymisation.json
HarrisonBoyleThomas: I initially thought so, but it turns out AIEDocument doesn't have any PII so it was dropped

HarrisonBoyleThomas: It only has a `path` column in purview, but i've confirmed this is incorrectly tagged

odw/core/io/synapse_file_data_io.py
HarrisonBoyleThomas: Updated

workspace/pipeline/pln_historical_anonymisation.json
KalyaniNik: ok thanks for confirmation

KalyaniNik (approved):
LGTM

PR #2646, opened 2026-07-22
Updating pipeline: rel_34_0_0_post_deployment

HarrisonBoyleThomas:
Ticket: https://pins-ds.atlassian.net/browse/THEODW-3176

While resolving some bugs in the modular pipeline i noticed that `rel_34_0_0_post_deployment` didn't seem to have been run in test/prod. I also noticed that the first stage of this pipeline will delete a lot of important entries (which doesn't seem to have been done in dev), so I have removed this stage. The pipeline now just renames some values, and adds a new entry


KalyaniNik (approved):
LGTM

